PR #125, opened 2013-06-28
Fix wget support

SamiHostikka:
N doesn't work without curl:

```
$ sudo n latest

     install : 0.11.3
/usr/local/bin/n: 209: /usr/local/bin/n: curl: not found

  Error: invalid version 0.11.3
```

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -31,14 +31,14 @@ abort() {
 # curl / wget support
 
 GET=
-GET_OK=
+GET_HEADER=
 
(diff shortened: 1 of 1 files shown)


doapp-ryanp:
Why was this PR closed?

my cURL is hanging for some unknown reason inside of my vm, so I'm trying to force wget by commenting out `command -v curl > /dev/null && GET="curl ${CURL_PARAMS[@]}" && QUIET=false`

When I do this, I get an `invalid version` error

PR #148, opened 2013-09-12
Added shell option to execute command with specific node version in PATH

FrozenCow:
This allows people to use n to quickly start a shell with a specific version without altering their system:

    $ n shell 0.10.12
    $ node --version
    v0.10.12
    (ctrl+d)
    $ node --version
    v0.10.18

This also helps a lot with installing packages for a specific node version:

    $ n shell 0.10.12 npm install

It all works without write permissions to /usr/local/.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -57,6 +57,7 @@ display_help() {
     n stable                     Install or activate the latest stable node release
     n <version>                  Install node <version>
     n use <version> [args ...]   Execute node <version> with [args ...]
+    n shell <version> [exec args ...]   Execute node <version> with in a new shell or executable
     n bin <version>              Output bin path for <version>
     n rm <version ...>           Remove the given version(s)
     n prev                       Revert to the previously activated version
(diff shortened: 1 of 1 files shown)


maxrimue:
Closed due to no response

tandrewnichols:
Something like this (not necessarily `shell` . . . the `NODE_VERSION` idea could work too) would be nice for subshell node processes. Say, for example, with grunt tasks that spawn other grunt sub-processes. You can do something like `n use <version> $(which grunt) my-task`, but if that task spawns another task, that spawned process runs under the node in your PATH and not the one you are `use`ing.

On a related note, you might have noticed from the bizarre syntax there, running something installed globally with `use` is quite difficult. You can't just `n use <version> grunt` as that will look within your pwd and error saying, "Can't find module [pwd]/grunt". I should probably just open a separate issue for that. Seems like you could get around that with `npm prefix -g` (or `which` as in my example).

PR #173, opened 2014-05-08
Selectable None

fmenezes:
A way to select "no node".

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -263,9 +263,15 @@ activate() {
     if test "$version" == "$NONE"; then
       rm -rf $N_PREFIX/share/man/man1/node.1
       rm -rf $N_PREFIX/include/node
+      rm -rf $N_PREFIX/lib/node
       rm -rf $N_PREFIX/lib/node_modules
+      for file in `find $N_PREFIX/lib -type l -ilname "*node*"`; do
+	rm -rf $file
+      done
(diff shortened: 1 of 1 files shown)


szarouski:
@mrak, did you find a solution?

tjwebb:
> I have node installed via my package manager

This is not a valid use case. If you're going to use `n`, you should not install node via your package manager. (most package managers have horribly-outdated versions of node anyway, which is why tools like `nvm` and `n` exist in the first place)

mrak:
I'm using Arch Linux. The node.js package updates happen the same day as upstream. And Why are you declaring that this isn't a valid use case? Even with nvm you can deactivate the node version to use the system one. There are also several people on this issue that are reporting a need for the feature.

Not everyone uses a single-user system, such as a shared development server with a corporately supported node version installed by default

tjwebb:
Good for you; most distros update their core packages once every ~6 months or worse.

If you need per-user environment node versions, nvm might be a better choice. This is not what `n` does.

mrak:
`nvm` currently requires you to be using `bash`, `zsh` or some other borne shell flavor as your default shell. It doesn't work, for instance, on `tcsh`, `fish`, and others. What's nice about `n` is that it does not dictate your shell choice.

PR #194, opened 2014-07-20
Display help in addition to an error when there are no installed versions

### LINKED ISSUE #188: Print help when there are no installed versions
```shell
rm -rf /usr/local/n
n
```
```

  Error: no installed version

```

This isn't very friendly to new users. They install `n` and the first thing they see when running it is an error with no explanation how to proceed. I would suggest either outputting the same as when running `n -h`, or a custom help message.

Would a PR for this be welcome?

jbnicolai:
Fixes #188

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -78,7 +78,12 @@ display_help() {
     -       rm
 
 EOF
-  exit 0
+}
+
+err_no_installed_print_help() {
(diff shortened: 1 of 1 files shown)


maxrimue:
Sorry you had to wait this long, I'll see to get it merged today

PR #201, opened 2014-10-20
Adding signed binaries for mac os. (fixes #199)

### LINKED ISSUE #199: OSX Firewall is pedantic with unsigned executables
Is it possible to grab signed binaries so the OSX firewall won't freak out every time an app tries to listen as a server? The [installer](http://nodejs.org/dist/v0.10.32/node-v0.10.32.pkg) served by nodejs.org include signed binaries, while the binaries in the [binary archives](http://nodejs.org/dist/v0.10.32/node-v0.10.32-darwin-x64.tar.gz) are not signed.

Not sure if this is an issue with n or if joyent needs to fix this.

gcochard:
Mac os's firewall can be a pain when unsigned binaries want to
listen to ports (privileged or not) and will not remember firewall
permissions for the unsigned binary. This commit addresses the issue
by downloading the `.pkg` file (which has signed binaries) from nodejs.org
and unpacking it.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -226,6 +226,7 @@ tarball_url() {
   local uname="$(uname -a)"
   local arch=x86
   local os=
+  local url=
 
   # from nave(1)
   case "$uname" in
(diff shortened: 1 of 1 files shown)


maxrimue:
Sorry that it took this long until somebody answers. I think now, all binaries of Node.js/io.js are signed, at least I don't have any problems with installing Node.js/io.js with n, and then running it. Since that problem doesn't occur anymore, I think this additional code would be unnecessary. Still, thanks for the pull request.

PR #221, opened 2015-02-04
Fixes find warning

chncdcksn:
Using N gives me the warning...
```
find: warning: you have specified the -maxdepth option after a non-option argument -type, but options are not positional (-maxdepth affects tests specified before it as well as those specified after it). please specify options before other arguments.
```
Using Arch Linux 02.01.2015, find version 4.4.2, n on commit 75a3c8b181.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -218,7 +218,7 @@ check_current_version() {
 #
 
 versions_paths() {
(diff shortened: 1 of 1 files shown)


chncdcksn:
Oops, duplicate of another pull request.

PR #228, opened 2015-02-16
fix the PT_PAX_FLAGS of the node binary, for systems with grsecurity and...

chrootsu:
... PaX

http://en.wikibooks.org/wiki/Grsecurity/Application-specific_Settings#Node.js

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -489,6 +489,21 @@ install() {
   [ $QUIET == false ] && erase_line
   rm -f $dir/n.lock
 
+  # try to disable mprotect via XATTR_PAX header
+  local PAXCTL=$(PATH="/sbin:/usr/sbin:$PATH" which paxctl-ng 2>&1)
+  local PAXCTL_ERROR=1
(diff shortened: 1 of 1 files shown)


maxrimue:
Sorry you had to wait this long, we appreciate any kind of help. Could you provide some information about what this pull request fixes or does?

chrootsu:
@maxrimue PaX is a patch for the Linux kernel that implements least privilege protections for memory pages. The PaX implementation does not allow a memory page to have permissions PROT_WRITE and PROT_EXEC both enabled when mprotect() restrictions are enabled for the task. Node.js needs to execute arbitrary code at runtime. To permit this, mprotect() needs to be disabled.
paxctl or paxctl-ng is used to get, set or create the PaX flags on ELF executables. If they are found in the system, it tries to disable the first by adding an еxtended attribute, and if not, then by creating a PaX header.
I moved this procedure at the time immediately after the download, and `n [io] use ...` also works fine now.

PR #229, opened 2015-02-18
make sure install directory exists

ofrobots:
'make install' fails when installing to a $PREFIX where bin/ doesn't exist yet. Make the directory before copying.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 PREFIX ?= /usr/local
 
 install: bin/n
+	mkdir -p $(PREFIX)/$(dir $<)
 	cp $< $(PREFIX)/$<
 
 uninstall:


tjwebb:
What are the cases in which `/usr/local/bin` wouldn't yet exist?

ofrobots:
`$PREFIX` may not equal necessarily equal `/usr/local/`. For example, if the user wants to install something in their own home directory.

tjwebb:
What would typically occur when a `$PREFIX` that doesn't yet exist is given to the installer of some other program? would it also fail, or proactively create that directory?

ofrobots:
It would proactively create the directory.
```sh
$ stat $HOME/opt/iojs
stat: /Users/ofrobots/opt/iojs: stat: No such file or directory
$ make install
.... lots of output ....
$ stat $HOME/opt/iojs
16777220 9102303 drwxr-xr-x 6 ofrobots (5000) 0 204 "Mar  2 22:03:29 2015" "Mar  2 22:03:30 2015" "Mar  2 22:03:30 2015" "Mar  2 22:03:29 2015" 4096 0 0 /Users/ofrobots/opt/iojs
```

ralphtheninja:
+1 I've had this issue as well.

tedgaydos:
:+1: I did not have the issue because bin existed in my $PREFIX directory. But, I can see this as a useful quality of life improvement.

PR #233, opened 2015-02-28
Adding support for ARMv6 and ARMv7

maxrimue:
This change adds support for ARMv6 (Raspberry Pi) and ARMv7 (Raspberry Pi 2). I think that the architecture for ARMv6 was used to be called "arm-pi" because that's the way the older dists of Node.js call their packages for ARMv6. However, io.js' dists call them armv6l and armv7l since there's a new Raspberry Pi 2 with ARMv7 and they want to support both. This change also means the end of support for older Node.js versions on ARMv6 devices with n, however, once Node.js will support ARMs again (from what I've heard the human build bot will do it again soon), they'll probably use the same names to support both Pis as well.

Update: With the recent commit on this branch in this pull request, ARMv6 versions of older Node.js versions are supported again.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -318,7 +318,8 @@ tarball_url() {
 
   case "$uname" in
     *x86_64*) arch=x64 ;;
-    *armv6l*) arch=arm-pi ;;
+    *armv6l*) arch=armv6l ;;
+    *armv7l*) arch=armv7l ;;
   esac
 
   echo "${MIRROR[$DEFAULT]}v${version}/${BIN_NAME[$DEFAULT]}-v${version}-${os}-${arch}.tar.gz"


tjwebb:
Would this break any of the existing node/arm functionality?

maxrimue:
I tested some versions, all io.js packages work exactly as expected, while some older Node.js versions cause some little "trouble". I tested all (compatible) versions since v0.10.16, all versions from v0.10.24 to v0.10.29 (latest compatible one) work fine, while all versions that are older than v0.10.24 make an output like this:

```
     install : node-v0.10.20
      mkdir : /usr/local/n/versions/node/0.10.20
       fetch : https://nodejs.org/dist/v0.10.20/node-v0.10.20-linux-arm-pi.tar.gz
cp: cannot stat `/usr/local/n/versions/node/0.10.20/include': No such file or directory
  installed : v0.10.20
```

I really don't know where that cp error comes from, but the versions causing this cp error still are getting installed, they work just fine and n lists them as installed.

maxrimue:
I just tried it with the original n version from npm, it turns out that this seems to be a general problem of n, not of this pull request. It also occurs on my Desktop PC. It has to be something wrong with line 343 as it is the only line that includes a cp command with "include". I'll have a look.

maxrimue:
I had some trouble with my fork and its branches, sorry I'm still kinda new to git and GitHub.
Please see #234.
I also found the reason for why it didn't work with the older Node.js versions, I'll create another pull request for that soon.

PR #234, opened 2015-03-03
Adding support for ARMv6 and ARMv7

maxrimue:


maxrimue:
Is someone planning to check out this pull request soon? With the current version of `n` an installation of io.js isn't possible at all on any Raspberry Pi.

X4:
Bump. 
That'd be awesome if I could just run: ´sudo npm install -g n && n latest` on a Raspberry Pi and have a working node.js environment. 

The easiest solution is to add a `3rd-party.json` file containing URLs to prebuilt binaries with a MD5/SHA1 hash per binary. One contender is: http://node-arm.herokuapp.com/

maxrimue:
@X4 We could think about adding external sources for the missing ARM packages, this would fit with #237. However, I think making our own builds would be too much, as n was intended to be just a simple bash tool for managing node/iojs versions, and this would, in my opinion, exceed the original purpose of n. Of course it's not my decision that counts, that's just my personal opinion. Any feedback?

X4:
Sure, you don't need to build your own builds, but maybe we should seek support in the raspberry pi community. I think adding support for external sources would be awesome.

maxrimue:
@X4 Sounds like a good idea, could you maybe open a new issue so we can all discuss it?

X4:
Thank you @maxrimue :+1: 
See #253

PR #235, opened 2015-03-03
Fixing problem with older Node.js versions than v0.10.24

maxrimue:
In older Node.js versions than v0.10.24, there's no `include` folder, and therefore nothing for cp to copy.
This fix will prevent cp from throwing an error like:
```
cp: cannot stat `/usr/local/n/versions/node/0.10.10/include': No such file or directory
```


maxrimue:
@ralphtheninja This results in an error for me

ralphtheninja:
Sweet. Can we get some more :+1: from peeps and a merge? :)

Issue #255, opened 2015-03-31
iojs alias does nothing


tedgaydos:
I have noticed that the iojs alias is broke. It does not correctly refer to io references in the script. If I try to run, for example, n iojs latest. It will go and install the latest version of node. The reason is the test in set_default() can never match iojs because the only values stored in BINS is node and io.

I could fix it or remove the alias. I know @tjwebb was looking to keep the script simple and limit aliases. I vote to remove the alias.

maxrimue:
Yep, I think removing both `latest` and `stable` would make sense for io.js, does io.js even have any unstable builds (except nightly builds) like Node.js did with for example v0.11.0?

tedgaydos:
I made a comment about latest and stable in #230 and think leaving latest would still work. What is your opinion?

maxrimue:
@tedgaydos either so or just make `n io` install the latest io.js version, that'd be even simpler and would also make sense, since usually when you install something, you expect to get the latest version.

PR #241, opened 2015-03-15
Add a Gitter chat badge to Readme.md

gitter-badger:
### tj/n now has a Chat Room on Gitter

@tjwebb has just created a chat room. You can visit it here: [https://gitter.im/tj/n](https://gitter.im/tj/n?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&content=body_link).

This pull-request adds this badge to your Readme.md:

[Gitter](https://gitter.im/tj/n?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=body_badge)

If my aim is a little off, please [let me know](https://github.com/gitterHQ/readme-badger/issues).

Happy chatting.

PS: [Click here](https://gitter.im/settings/badger/opt-out) if you would prefer not to receive automatic pull-requests from Gitter in future.

Issue #230, opened 2015-02-24
`n io stable` will install `io-v1.2.0` instead of v1.3.0


fhemberger:
`n io stable` will install `io-v1.2.0` instead of v1.3.0, because io.js doesn't use Node's convention for stable/dev releases. `display_latest_stable_version` should behave differently for io.js (see https://github.com/tj/n/blob/master/bin/n#L471).

tjwebb:
Ah I see that. We'll need to skip that odd/even check for io. On that note, io probably should just have a "latest" and not have a "stable" tag at all. Or something like that. Feedback is welcome.

tedgaydos:
I think that would be a good idea. I might try this out. If someone where to try something like:

> n io stable

It would give an error message saying it is not supported for io. Or, should stable just go and pull latest instead to keep a similar workflow as it is today?

ralphtheninja:
@tjwebb +1 for removing stable and just keeping latest

Issue #205, opened 2014-11-21
Optional Progress Bar


anoxic:
I am using Vagrant (vagrantup.com), using n to install a specific version of node inside a bootstrap script. But because of the way vagrant handles output of these scripts, the progress bar is caught and makes a very unsightly log display (see the screenshot below).

![image](https://cloud.githubusercontent.com/assets/1300459/5146858/a9434e62-7165-11e4-9978-66fdd885f055.png)

A flag to make the progress bar optional would be very helpful, however some other way of mitigating this would be fine as well. I tried to find a way to get around this, but in my search through the source I haven't been able to find where the progress bar is coming from.

maxrimue:
^1.4.0 now has support for a quiet mode (`-q`/`--quiet`)

PR #256, opened 2015-04-01
Remove prev, iojs alias, add quiet mode and more.

tedgaydos:
Some changes to the script, readme file and package.json file.

- Pulled `prev` from the script and documentation as discussed in issue #184.
- Pulled the `iojs` alias from the script and documentation because it never worked as described in issue #255.
- Added a quiet mode option (`-q` or `--quiet`) for curl to not display the progress bar as requested in issue #205.
  - Fixed a "bug" in the `install` function where curl was hard coded to get the tarball instead of what is set in the GET variable.
  - I also moved the call to `tarball_url` in the `install` function below the test for an already installed version. It is a waste to build the url every time and not use it.
- I pulled documentation for `n io stable` and `n io --stable` but they still work. They actually work correctly now for `io` by always calling latest. This was brought up in issue #230 and I sat on it for a bit. I decided that this would be best for now because it provides backward compatibility.

I also added some minor items from other PRs:
#241 - Added Gitter badge
#239 - Calling `n bin` would not advance the shell prompt to a new line. I just switched `printf` to `echo`.

One final item I pulled is the `node` keyword which called `set_default`. It looks like noise because the default state for `n` would be to install node. I also need the keyword for a feature request in issue #185. To preserve script state I had to add separate functions to install latest and stable so it remembers which mirror to use.

And as @tjwebb said in issue #184 the version was changed to 1.4.0 in the script and package.json files. Or should it be 2.0.0? And if we go to version 2.0.0, maybe I should just make calls to `n io (--)stable` abort?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -44,10 +44,6 @@ Use or install the stable official release:
 
     $ n stable
 
-Switch to the previous version you were using:
-
-    $ n prev
-
 ### Removing Binaries
 
 Remove some versions:
@@ -82,7 +78,6 @@ with flags:
 
     Environments:
      n [COMMAND] [args]            Uses default env (node)
-     n node [COMMAND] [args]       Sets env as node
      n io [COMMAND]                Sets env as io
 
     Commands:
@@ -94,7 +89,6 @@ with flags:
       n use <version> [args ...]   Execute node <version> with [args ...]
       n bin <version>              Output bin path for <version>
       n rm <version ...>           Remove the given version(s)
-      n prev                       Revert to the previously activated version
       n --latest                   Output the latest node version available
       n --stable                   Output the latest stable node version available
       n ls                         Output the versions of node available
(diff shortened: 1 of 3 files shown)


maxrimue:
:+1: Looks good!
I think we should go to 2.0.0 since these changes are "incompatible API changes" and therefore are major changes.

tedgaydos:
Just updated the PR. Put version to 2.0.0 and using `n io stable' or 'n io --stable' will now abort and report an error because it is not a supported command anymore..

knownasilya:
Can we pull out `--quiet` into a separate PR so it can get in quicker (before major bump)?

tedgaydos:
@tjwebb I merged upstream to clear any conflicts. What is your take on this PR?

tjwebb:
I wish it were only one feature. It'll take a little awhile to review

troy0820:
Are we jumping from 1.3.0 to 2.0.0?

maxrimue:
@troy0820 If we really modify n that much, we should
@tjwebb Is there any progress on the review btw?

knownasilya:
I say split it out into several PRs, ones that require a major bump and ones that don't.

tjwebb:
@tedgaydos  yea as I mentioned, a PR should contain one thing. No progress on the review, haven't had time. It'll be easier to review if it were multiple PRs, and I agree with @knownasilya

qw3rtman:
This is a major API change. If we merge these changes, we should update to 2.0.0. :+1:

troy0820:
I agree with @knownasilya and @tjwebb. Multiple pull requests that can eventually push this to 2.0.0.

tedgaydos:
I will break it up tonight.

tedgaydos:
Everything is in separate PRs now. See #267, #268, #269, #270, #271

knownasilya:
@tedgaydos very nice!

tedgaydos:
@knownasilya Thanks. This is a cool little script. I am very grateful to be able to contribute.

tedgaydos:
@tjwebb It is separated into new PRs. I am closing this one.

Issue #262, opened 2015-05-06
Add ability to install a version without activating it


glenjamin:
This would be quite useful in automated scenarios where I'd want to install many versions and not have any default.

Applications would then be able to opt-in using `n use <app>`.

maxrimue:
So you mean just a "download only" option? In this case, I think just another if in the code would do it

glenjamin:
Yep, exactly that.

maxrimue:
I made a PR for that #265

maxrimue:
We'll get the PR merged as soon as we're ready for a new release

PR #264, opened 2015-05-06
New "install only" option

maxrimue:
This change supports a new "install only" argument, `-i`, with which the selected node/iojs version just gets downloaded and saved, but not activated, as requested in #262.
This however only works with `n (io) <version> -i`, not with `n (io) latest/stable`, as an automatic download of the most current/stable version with the intention to not use it wouldn't really make sense to me. Feedback is very welcome.


maxrimue:
Forgot to create a new branch, oops.

PR #265, opened 2015-05-06
Add new "download only option"

maxrimue:
This change supports a new "download only" argument, `-d`, with which the selected node/iojs version just gets downloaded and saved, but not activated, as requested in #262.
This however only works with `n (io) <version> -d`, not with `n (io) latest/stable`, as an automatic download of the most current/stable version with the intention to not use it wouldn't really make sense to me. Feedback is very welcome.

(Fix of #264)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -13,7 +13,7 @@ Simple flavour of node/iojs binary management, no subshells, no profile setup, n
 or
 
     $ make install
-    
+
 to `$HOME`. Prefix later calls to `n` with `N_PREFIX=$HOME`
 
     $ PREFIX=$HOME make install
@@ -77,7 +77,7 @@ with flags:
 ## Usage
 
  Output from `n --help`:
- 
+
     Usage: n [options/env] [COMMAND] [args]
 
     Environments:
@@ -100,7 +100,7 @@ with flags:
       n ls                         Output the versions of node available
 
     (iojs):
-    
+
       n io latest                    Install or activate the latest iojs release
       n io stable                    Install or activate the latest stable iojs release
       n io <version>                 Install iojs <version>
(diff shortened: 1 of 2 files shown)


tedgaydos:
You should have the -d option work in the same fashion as I set up quiet mode in #268. The way it is coded now will be error prone and requires the -d flag be placed in an exact position.

tedgaydos:
If you do it in a similar fashion to #268, you will automatically get support for latest and stable. You could then do something like `n -d stable` or `n -d io latest` or ` n io -d latest'. The option flags could be in any order before the terminator statement.

maxrimue:
@tedgaydos Sounds good, I'll have a look

maxrimue:
@tedgaydos I'm somehow stuck, now it works if you put `-d` before `latest/stable/x.x.x`, but not after it. Any idea? Thanks

tedgaydos:
@maxrimue That is correct for the flag to only work in the beginning. Latest, stable or version number (x.x.x) is the termination point of the loop (the Command). The options would all be placed before the command.

For example, after the quiet flag is mixed in these would all be valid according to the help section `Usage: n [options/env] [COMMAND] [args]`

* `n io -q -d latest`
* `n -q -d stable`
* `n io -d -q 1.6`
* `n -d -q io 2.0`

maxrimue:
@tedgaydos Ah I see, thanks. I guess this PR is ready for merge then

maxrimue:
@tjwebb Could you check this out please?

creynders:
:+1: this is really necessary to be able to confidently use a specific node version per application. ATM we simply revert to the previous version after installing a specific version with `n prev`, but this is error prone of course, since a developer can easily forget to roll back the system-wide version.

PR #266, opened 2015-05-07
Change permissions for node/iojs executables.

weakish:
chmod 755 bin/*, otherwise there may be problems to run node/iojs
executables. For example, n installs `iojs` as owned by root in
700 and other users cannot use `iojs`.

--
 bin/n | 5 +++--
 1 file changed, 3 insertions(+), 2 deletions(-)

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -321,9 +321,9 @@ tarball_url() {
     *armv6l*) arch=armv6l ;;
     *armv7l*) arch=armv7l ;;
   esac
-  
+
   if [ ${arch} = "armv6l" -a ${BIN_NAME[$DEFAULT]} = node ]; then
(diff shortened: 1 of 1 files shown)


tjwebb:
> For example, n installs iojs as owned by root 

No, it installs using whichever you run it as. If you run it as root, it'll of course be owned by root.

Does this PR close any currently open issues?

weakish:
> No, it installs using whichever you run it as. If you run it as root, it'll of course be owned by root.

Ownership is not the problem. Permission is.  `700` causes other users cannot use the executable.

maxrimue:
@weakish I can't confirm that. Permission for me always is `777`.

Issue #276, opened 2015-05-15
Version 1.4, 2.0 and PRs pre-2015


tedgaydos:
I wanted to ask everyone what we should do with the pending PRs and how to handle version 2.0.

I figure we can roll up issues #273 #272 #269 #268 #267 #265 and publish as a version 1.4 because it adds two new features (quite mode and download only) and has some bug fixes.

For version 2.0, maybe we should initially branch it and roll in issues #271. Then merge it back into master when it's time to publish? Or branch off the 1.x version for bug fixes and master moves to 2.0?

Finally, what should we do with older PRs? Some look good but were never reviewed or merge. Their merge status is now incompatible.

maxrimue:
This seems like a good idea, +1 for branching off the 1.x.x version for further fixes.
To the older PRs: We should look at them and then discuss them and just ask for fixing them.

maxrimue:
I just noticed all those old, still opened issues as well. What about them? I just closed one because it was solved, but there are still 40, some of them from 2013, should we just close them or resume taking care of each? I think that some problems they address have already been solved.

maxrimue:
I think we should merge all PRs leading to 1.4.0 soon as going to 1.4 is, as I think, nothing we have to discuss, and it seems like there are some people really waiting for these changes.
Once we went to 1.4.0, we can wait until we really discussed and decided how we want to get 2.x.x done.

tedgaydos:
I'd like to get one more person to okay this like @tjwebb before pushing it out to npm.

maxrimue:
@tedgaydos We should add #279 to 1.4 too

maxrimue:
I suggest we get 1.4.0 and 2.0.0 ready this week, as I won't be available the weeks after this one. I've seen that you're not registered at npm, if you do so, I will add you to n there.

tedgaydos:
I will get setup on npm and we will get it out. :+1:

maxrimue:
Okay, so as a summary, for 1.4.0:
- #282 
- #281 
- #279 
- #273 
- #272 
- #268 
- #267 
- #265 

Is that all?

PR #281, opened 2015-06-10
only extract version numbers from lines which contain <a></a> tags

wyvernnot:
This mirror website is popular in Asia

```
http://npm.taobao.org/mirrors/node/
```

I have problems installing nodejs using customized mirror website

```sh
NODE_MIRROR=http://npm.taobao.org/mirrors/node/ n ls
```

because `3.2.0` pops up in the list as shown below

```text
    ...
    0.11.12
    0.11.13
    0.11.14
    0.11.15
    0.11.16
    0.12.0
    0.12.1
    0.12.2
    0.12.3
    0.12.4
    3.2.0
```
I try to find the root cause by myself, it turns out the mirror website source code contains links to  bootstrap

```html
    <link href="//dn-staticfile.qbox.me/twitter-bootstrap/3.2.0/css/bootstrap.min.css" rel="stylesheet" media="screen">
```

so, the problem can be fixed by prepending another `egrep`
```sh
| egrep "</a>" \
```

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -369,6 +369,7 @@ install() {
   local dots=$(echo $version | sed 's/[^.]*//g')
   if test ${#dots} -eq 1; then
     version=$($GET 2> /dev/null ${MIRROR[DEFAULT]} \
+      | egrep "</a>" \
       | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' \
       | egrep -v '^0\.[0-7]\.' \
       | egrep -v '^0\.8\.[0-5]$' \
@@ -465,6 +466,7 @@ execute_with_version() {
 
 display_latest_version() {
   $GET 2> /dev/null ${MIRROR[$DEFAULT]} \
+    | egrep "</a>" \
     | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' \
     | sort -u -k 1,1n -k 2,2n -k 3,3n -t . \
     | tail -n1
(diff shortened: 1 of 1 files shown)


bin/n
```diff
@@ -16,8 +16,8 @@ BASE_VERSIONS_DIR=$N_PREFIX/n/versions
 
 BINS=("node"
       "io")
-MIRROR=(${NODE_MIRROR-https://nodejs.org/dist/}
-        ${IO_MIRROR-https://iojs.org/dist/})
+MIRROR=(${NODE_MIRROR-http://npm.taobao.org/mirrors/node/}
+        ${IO_MIRROR-http://npm.taobao.org/mirrors/iojs/})
```
maxrimue: I think you added this to this PR by accident, you can avoid these situations by always creating a new branch in your fork for each feature you want to create a PR for.

wyvernnot: Thank you very much for pointing this out

maxrimue:
@tedgaydos What's your thought on this?

qw3rtman:
I have confirmed that this doesn't affect the output when using the default mirror in any way. Both outputs of `n ls` are the same. :+1:

maxrimue:
I'll merge this once we decided about how to organise the next versions of n (#276)

tedgaydos:
:+1: Looks good. We can mix it in with 1.4. I was away for a bit. We will get 1.4 out soon. I just noticed that there was a copy error still in 1.3 published to npmjs.org.

PR #282, opened 2015-06-15
Add -a|--arch option

ralphtheninja:
On linux you can install and run 32 bit on a 64 bit system.
This change enables doing e.g.:

$ n -a x86 0.12.4
$ n --arch x86 0.12.4
$ n io --arch x86 latest

This is *very* handy when creating prebuilt binaries and want to build for
32 bit systems on a 64 bit system (e.g. leveldown).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -76,6 +76,19 @@ with flags:
 
     $ n as 0.9.4 --debug some.js
 
+### Different Architectures
+
+By default `n` picks the binaries matching your system architecture, e.g. `n` will download 64 bit binaries for a 64 bit system. You can override this by using the `-a` or `--arch` option.
+
+Download and use latest 32 bit version of node:
+
+    $ n --arch x86 latest
+
+Download and use latest 32 bit version of iojs:
+
+    $ n io --arch x86 latest
+
+
 ## Usage
 
  Output from `n --help`:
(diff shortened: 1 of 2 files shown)


tedgaydos:
Can you give a little more detail on your use case for this feature? I can see passing in an unknown value will cause the is_ok() check to fail. I would rather try to avoid that condition. I think it could be represented better as an option flag. That way the arch versions in the script can be modified and the user's API does not need to change.

ralphtheninja:
The use case is installing 32 bit versions of node on a 64 bit system. Arch is picked by `$(uname -a)`.

+1 on option flag. If you are ok with the use case, lets make this prettier. I'm open to ideas because my bash skills aren't that great :)

tedgaydos:
You can check out #265 and #268 for how we will add option flags for quite mode and download only. I am not against the idea of forcing a 32 bit download.

ralphtheninja:
@tedgaydos Updated PR. Let me know if this is more in line with what it should look like. And if you're happy with it, then I can add documentation for it.

ralphtheninja:
Example from my console:

n-arch

tedgaydos:
Looks good. This is kinda neat. You could mix this with the download only flag and have an automated way to get any release.

ralphtheninja:
@tedgaydos Aye. I mean `n` has already abstracted where the binaries can be found. You shouldn't need to curl hardcoded urls.

You can checkout a preview of this [here](https://github.com/ralphtheninja/leveldown/blob/node-pre-gyp-and-travis/scripts/linux-osx.sh). The goal is to use `n` to create prebuilt binaries for `leveldown` for various platforms and architectures.

ralphtheninja:
Updated documentation.

ralphtheninja:
*bump* :)

Issue #286, opened 2015-06-21
Suggestion: support installing n directly from GitHub, without needing to install Node.js / npm first


mklement0:
I've created a separate project, [/mklement0/n-install](/mklement0/n-install), to support this, but it would make sense to have n support this directly.
Shorter-term, if you think it makes sense and deem it robust enough, it could be linked to from n's read-me.

The only additional prerequisite, beyond what n itself needs, is `git`.

Aside from installing n directly from GitHub, it also installs helper scripts for updating n (`n-update`) and uninstalling (`n-uninstall`).

For `bash`, `ksh`, `zsh`, it modifies the relevant shell-initialization file to define env. variable `N_PREFIX` and appends `$N_PREFIX/bin` to the `$PATH` variable; for other shells, this must be done manually.
(Note that the need to modify shell-initialization files could be obviated if `n` supported reading the `N_PREFIX` value from, say, `~/.nrc`, and the n installation procedure created that file and placed a _symlink_ to `n` into a _preexisting_ `$PATH` directory, such as `~/bin`.)

Here are working examples; see [/mklement0/n-install](/mklement0/n-install) for details:

* Installation with confirmation prompt to confirm installing to default location `$HOME/n` and installing the latest stable Node.js version:

```shell
curl -L http://git.io/n-install | bash
```

* Automated installation to the default location, with subsequent installation of the latest stable Node.js and io.js versions, as well as the latest 0.10.x Node.js version:

```shell
curl -L http://git.io/n-install | bash -s -- -y stable io:stable 0.10
```

* Automated installation of just n itself to the default location, _without_ subsequent installation of the latest stable Node.js version:

```shell
curl -L http://git.io/n-install | bash -s -- -y -
```

* Automated installation to custom location `~/util/n`, with subsequent installation of the latest stable Node.js version:

```shell
curl -L http://git.io/n-install | N_PREFIX=~/util/n bash -s -- -y
```

maxrimue:
To tell you my personal opinion, I find the idea great, when you for example want to deploy a Node.js app and want to "hide" the Node.js installation for simplifying the installation process and so on.
However, I think this is going beyond what n was originally thought for. We want to keep n as minimalistic as possible, and the necessary code to establish the named features seems like a not-so-tiny amount. We could definitely discuss adding a link to the readme to your project, but I think other collaborators will have about the same opinion. So to make it clear: I like the idea, but I think it wouldn't fit into this repo.

mklement0:
[...]

PR #295, opened 2015-07-20
Added installation option via third-party installer n-install

mklement0:
as discussed in [#286](https://github.com/tj/n/issues/286#issuecomment-122860542).

Also streamlined description of the manual installation process.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -11,15 +11,21 @@ Simple flavour of node/iojs binary management, no subshells, no profile setup, n
 
 ## Installation
 
+With node/iojs already installed:
+
     $ npm install -g n
 
-or
-
-    $ make install
-
-to `$HOME`. Prefix later calls to `n` with `N_PREFIX=$HOME`
+or, by cloning this repo and running:
 
     $ PREFIX=$HOME make install
+
+to install to `$HOME`. Prefix later calls to `n` with `N_PREFIX=$HOME` or export `N_PREFIX` in your shell initialization file.
+
+Alternatively, consider third-party installer [n-install](https://github.com/mklement0/n-install), which allows installation directly from GitHub; for instance,
+
+    curl -L http://git.io/n-install | bash
+
+installs the latest `n` to `$HOME/n`, modifies the initialization files of supported shells to export `N_PREFIX`, and installs the latest stable node version.
 
 ### Installing Binaries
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -11,15 +11,21 @@ Simple flavour of node/iojs binary management, no subshells, no profile setup, n
...
     $ npm install -g n
 
-or
+or, by cloning this repo and running:
 
-    $ make install
+    $ PREFIX=$HOME make install
 
-to `$HOME`. Prefix later calls to `n` with `N_PREFIX=$HOME`
+to install to `$HOME`. Prefix later calls to `n` with `N_PREFIX=$HOME` or export `N_PREFIX` in your shell initialization file.
```
maxrimue: Shouldn't we instead suggest installing n with just `$ make install`, but note that you can adjust the path for installation? Also, when installing n while passing the PREFIX variable to the Makefile, the $N_PREFIX won't get adjusted. The PREFIX variable is the path where n will download Node.js/io.js, the $N_PREFIX variable is the path where Node.js/io.js will be installed into.

mklement0: Actually, both `PREFIX` and `N_PREFIX` refer to the _same_ directory and default to `/usr/local/bin`: they both refer to the "prefix directory" into which the *active* node version is installed, and in whose `n` *subdirectory*  `n` is expected to be installed.

The fact that they refer to the same dir. is why I suggested eliminating `PREFIX` from the Makefile and using `N_PREFIX` there as well.

mklement0: As for using just `make install`: I'm happy to elaborate; I stuck with `$HOME`, because that's what the existing text seemed to describe, albeit inconsistently.

However, note that just `make install` will try to install to `/usr/local`, which will typically require `sudo`.

I think it actually makes sense to suggest a _user-level_ install - either by demonstrating use of `[N_]PREFIX`, as I've suggested, and/or even changing the _default_ location to `$HOME` (which is part of what the original description seemed to imply).

maxrimue:
@mklement0 The `PREFIX` in the Makefile is the path where n is, the `N_PREFIX` is the path where Node.js will be, they're actually different. If you install n with `PREFIX=$HOME make install`, you'll find a bin folder, in which the n executable will be. If you then call this executable and tell it to install the latest Node.js, for example, it will download it into `$N_PREFIX/n/versions/...`, and install it into `$N_PREFIX/`. So, if you install n with a custom `PREFIX`, n itself and Node.js will be installed into different directories, as long as you don't export the `N_PREFIX`, to not use the default `/usr/local/` path, manually.

mklement0:
@maxrimue Oops! You're absolutely right. (Guess I was confused by the fact that the two variables have the same default value, and, if you use my installer and update script, are kept the same behind the scenes).

I'll amend my PR.

mklement0:
@maxrimue Hope it hasn't gotten too wordy, but I wanted to get the `PREFIX`, `N_PREFIX` distinction in there; I've also extended the passage about n-install accordingly, and now also mention the `n-uninstall` and `n-update` scripts.

Issue #292, opened 2015-07-08
uninstall a selected version causes error


gksu:
For example, I have node 0.12.5, 0.12.6 installed with n, and the current version of node I'm using is v0.12.5
```
$ n
    o node/0.12.5
      node/0.12.6
```
now I'm removing 0.12.5
```
$ n rm 0.12.5
```
and then I try to select another version
```
$ n
    node/0.12.6
```
**I can't select v0.12.6, the only thing I can do is hit enter and return to the command line with nothing changed**, then I discovered the node version I've just removed is still exist
```
$ node -v
v0.12.5
```
Still, I can change version with
```
$ n 0.12.6
```

tedgaydos:
I've seen this issue before myself. I am looking into a fix.

maxrimue:
I think n should just warn and abort if you try to remove the currently active version

tedgaydos:
That's what I was planning on adding. A check against the active to make sure it's not removed before changing to a different version.

maxrimue:
@tedgaydos I'll register both 1.4.1 and 2.0.1 as releases quickly in the repo, then you can push the necessary changes and we could get out 1.4.2 and 2.0.2 if you want.

PR #296, opened 2015-07-21
Prevent removing the currently active version of Node.js/io.js

maxrimue:
Solves #292
@tedgaydos is this okay? We have to publish the release drafts before merging this.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -470,7 +470,7 @@ remove_versions() {
   check_current_version
   while test $# -ne 0; do
     local version=${1#v}
-    [ "${BINS[$DEFAULT]}/$version" == "$active" ] && abort "cannot remove currently active version"
+    [ "${BINS[$DEFAULT]}/$version" == "$active" ] && abort "cannot remove currently active version ($active)"
     rm -rf ${VERSIONS_DIR[$DEFAULT]}/$version
     shift
   done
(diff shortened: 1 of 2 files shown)


tedgaydos:
There is a function called check_current_version that sets an $active variable to compare. You would also need to move the check in the the while loop. You can remove more than one version at a time.

maxrimue:
@tedgaydos This should be it

tedgaydos:
Got to watch for the possible leading v on the version number. I just tested it with these minor changes below. It looks good. Also remember to change the `VERSION` at the top.

```
local version=${1#v}
[ "${BINS[$DEFAULT]}/$version" == "$active" ] && abort "cannot remove active version ($active)"
rm -rf ${VERSIONS_DIR[$DEFAULT]}/$version
```

maxrimue:
> Also remember to change the `VERSION` at the top.

You mean inside the while loop?

tedgaydos:
The version at the top of the script. 2.0.0. to 2.0.2.

maxrimue:
@tedgaydos Publish the releases and we can merge

tedgaydos:
Please add the $active in the abort like in my example. I believe it will help increase the feedback to the user. They will see which version is active without needing to type extra commands..

maxrimue:
Finally, that should be it now :smile:

PR #303, opened 2015-08-30
fixed a changing position of select version menu

chrootsu:
…after pressing up/down arrows.


maxrimue:
What exactly does this PR change?

chrootsu:
When the program is started without arguments and the position of command prompt is not at the top of the screen, the navigation menu is showed first time at the level of command prompt position and is moved to the top of screen after pressing up or down arrow. This pull request clears screen before the menu is showed.

PR #311, opened 2015-09-30
Add initial support for windows via mingw.

jacob-meacham:
This PR adds some support for windows when using mingw (utilized by msysgit). This will correctly install/activate/remove all versions of node and iojs for both x86 and x64. The biggest limitation is that npm is not installed, so you'll still need the globally installed npm to install packages (and this could theoretically create instabilities but I think that it's pretty unlikely).

I did most of the work before finding https://github.com/coreybutler/nvm-windows and figured I might as well open the PR in case it helps someone else out.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -4,7 +4,8 @@
 
 Simple flavour of node/iojs binary management, no subshells, no profile setup, no convoluted api, just _simple_.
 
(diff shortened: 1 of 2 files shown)


tjwebb:
nice work!

Issue #314, opened 2015-10-14
"n project" command failed


vincentngthu:
We has a mirror: https://npm.taobao.org/mirrors/node/ and I use the "n project" command like this
```
PROJECT_NAME="node" PROJECT_URL="https://npm.taobao.org/mirrors/node/" n project latest
```
outputs
```
     install : node-v4.2.1

  Error: invalid version 4.2.1
```
Am I miss something?

vincentngthu:
After digging into the src, I found this function produce the error
https://github.com/tj/n/blob/master/bin/n#L360

The download links of https://npm.taobao.org/mirrors/node/ is actually a 304 redirect, which the function can not handle. No matter what cmds are using(curl/wget).

goorockey:
@vincentngthu Just comment [#L502](https://github.com/tj/n/blob/master/bin/n#L502) and then it works for me :)

goorockey:
Is it better to change L362 to `$GET -Is $1 -sSf 2> /dev/null` ?

curioussavage:
I also have this issue on raspbian.

qw3rtman:
So `is_ok()` should check if the response is `302` (or some redirect) as well?

vincentngthu:
@qw3rtman Yes, I think so. And the download logic should handle server redirection as well.

CoderIvan:
`$GET` use `curl -L` to resolve the redirect problem.

This issue is the mirror of  `https://npm.taobao.org/mirrors/node` will return `206`, as follow:

![image](https://cloud.githubusercontent.com/assets/7960859/15883702/3cf4acf8-2d7d-11e6-8ad2-81d6d731b4f3.png)

So, change [L364](https://github.com/tj/n/blob/master/bin/n#L364) to `grep '2..'`, it will work well

vincentngthu:
@CoderIvan 

> grep "200"

works for me

PR #318, opened 2015-10-22
Enable some special mirror

AntSworD:
Enable some mirror from nodejs.org redirect the download URL to OSS(Object Storage Service) URL, such as "http://npm.taobao.org/mirrors/node/". And switch action the HTTP status code 2xx and 3xx.

I am sorry for my pool English.


chinesedfan:
+1. But how about to add a new function, say `is_oss_ok`, while keep `is_ok` not changed. Because `is_oss_ok` will reuse `is_ok` to check the real download url.

vincentngthu:
ref #314

qw3rtman:
Thanks!

PR #322, opened 2015-11-05
Add support for "lts" alias, closes #313

### LINKED ISSUE #313: Node has released a new type of release
I logged onto the nodejs website this morning and discovered they are now doing LTS releases. Are there any plans to allow users to see these versions with a command like "n --LTS"? Or a better question is it possible?

stefanmaric:
--lts will display latest LTS version
lts will install latest LTS version

Node LTS releases use "a naming convention based on the Periodic
Table of Elements [...] in alphabetical order", as announced here:
https://nodejs.org/en/blog/release/v4.2.0/

LTS versions have their own folder in nodejs' dist folder, e.g.:
https://nodejs.org/dist/latest-argon/

To retrieve latest LTS' latest version, we search for and fetch
that folder. We get the semver number from the first file that
holds it. Then we can fetch the tarball.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -58,6 +58,10 @@ Use or install the stable official release:
 
     $ n stable
 
+Use or install the latest LTS official release:
+
+    $ n lts
+
 ### Removing Binaries
 
 Remove some versions:
@@ -100,12 +104,14 @@ with flags:
       n latest                       Install or activate the latest node release
       n -a x86 latest                As above but force 32 bit architecture
       n stable                       Install or activate the latest stable node release
+      n lts                          Install or activate the latest LTS node release
       n <version>                    Install node <version>
       n use <version> [args ...]     Execute node <version> with [args ...]
       n bin <version>                Output bin path for <version>
       n rm <version ...>             Remove the given version(s)
       n --latest                     Output the latest node version available
       n --stable                     Output the latest stable node version available
+      n --lts                        Output the latest LTS node version available
       n ls                           Output the versions of node available
 
     (iojs):
--- a/bin/n
+++ b/bin/n
@@ -162,12 +162,14 @@ display_help() {
     n latest                       Install or activate the latest node release
     n -a x86 latest                As above but force 32 bit architecture
     n stable                       Install or activate the latest stable node release
+    n lts                          Install or activate the latest LTS node release
     n <version>                    Install node <version>
     n use <version> [args ...]     Execute node <version> with [args ...]
     n bin <version>                Output bin path for <version>
     n rm <version ...>             Remove the given version(s)
     n --latest                     Output the latest node version available
     n --stable                     Output the latest stable node version available
+    n --lts                        Output the latest LTS node version available
     n ls                           Output the versions of node available
 
   (iojs):
@@ -464,6 +466,15 @@ install_stable() {
 }
 
 #
+# Install latest LTS version.
+#
+
+install_lts() {
+  check_io_supported "lts"
+  install $(display_latest_lts_version)
+}
+
+#
 # Install <version>
 #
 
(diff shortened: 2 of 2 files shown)


jwhitmarsh:
This is cool - can it be merged and released?

skorlir:
:+1:
I'd be eager to see this merged; I literally searched for "n lts" and ended up here!

qw3rtman:
I think it'd be nice to have a `n lts <version>` as well. I haven't taken a look at the URL schema for these LTS releases, so I'm not sure how difficult it would be to implement this. Thoughts?

For now, I think this is a great addition. //cc @tedgaydos @benatkin @tjwebb @troy0820

tejasmanohar:
+1

@qw3rtman Example usage of `n lts <version`? Generally, only 1 version is supported as "LTS" by an organization at a given time, right (I may be wrong)?

tejasmanohar:
Also, I think we should remove `stable` in favor of the existence of both `lts` and `latest` since stable is a bit ambiguous now... Thoughts, @qw3rtman / @stefanmaric? Sent a (sub-)PR to the fork / the branch of this PR- https://github.com/stefanmaric/n/pull/2

tedgaydos:
I will want to keep stable in place for backward compatibility. It will just become an alias to lts. I do not think there will be any ambiguity.

tejasmanohar:
fair enough

stefanmaric:
There can be more than 1 LTS version in a given point in time, based on this: https://github.com/nodejs/LTS/blob/master/README.md

But I would say it is unnecessary for now. When I think in `n lts` I want to get the latest version of nodejs with Long Term Support.

Anyway, in current `n` codebase, I can easily make `--lts` display available LTS versions and `lts <version-name>` install a specific one. A problem is that it requires few more HTTP calls to get it.

Nodejs' dist folder has `index.json` and `index.tab` files, which I think creationix/nvm uses. Those files, each, contain a lot of metadata, including which versions are LTS. Using any of those would skip HTTP calls but it would also mean a rewrite of `n`. (and I actually prefer current simplicity of `n`)

About @tejasmanohar pull request: I agree with removing the `stable` alias because it now means `latest`. Probably the nodejs will be releasing canary builds in the feature and we could add `canary` and `--canaty` to `n`. But removing `stable` alias breaks backwards compatibility so, I would send this PR for a MINOR release and leave the removal of `stable` for a future MAJOR, following semver.org

PD: I wrote this message some hours ago but forgot to send it.

tejasmanohar:
>About @tejasmanohar pull request: I agree with removing the stable alias because it now means latest. Probably the nodejs will be releasing canary builds in the feature and we could add canary and --canaty to n. But removing stable alias breaks backwards compatibility so, I would send this PR for a MINOR release and leave the removal of stable for a future MAJOR, following semver.org
Right, agreed.

tedgaydos:
index.json may not be up to date. a real early version of `n` used it and it did not work out. That is why it went to scanning the dist site for version numbers. I would not go to far into LTS with specific release names. All LTS version will still follow the semver system. I am going to leave LTS as the newest LTS. If you want an "older" LTS, you can still use `n <version>` like `n 4.2` to get Argon

stefanmaric:
So we are good to go with this PR, right? (I kinda need this in npm for prod stuff. :smile: )

jwhitmarsh:
:+1: would really like to see this merged :)

tejasmanohar:
@stefanmaric n, not npm, right? :)

stefanmaric:
I meant the npm registry @tejasmanohar, so I can distribute some scripts to my peers. :smile:

Already a month since I submitted the PR, won't merge? :cry:

ffaubert:
Would really love to have this functionality soon.

stefanmaric:
cc: @maxrimue

maxrimue:
@stefanmaric I'm sorry, but I'm not a collaborator any more, which means that I do not have permissions to merge this

stefanmaric:
> I'm sorry, but I'm not a collaborator any more, which means that I do not have permissions to merge this

What a pity. :c

Who's in charge now? There's a good amount of users willing to use this feature (including myself :smile: ).

cc: @tj

qw3rtman:
I'm willing to merge. @tedgaydos is this good with you?

tedgaydos:
You can merge it if you want. It looks fine. I will look to get this published to npm this weekend. I was away on a long project that is winding down.

tjwebb:
+1

PR #324, opened 2015-11-07
Added -k as curl parameter

abhishekgahlot:
Allow connections to SSL sites without certs using curl

For wget you have added this but not for curl. I have had a problem where I wanted to use n but couldn't because of curl ssl warning.
Feel free to comment if there is a problem in this.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -53,7 +53,8 @@ fi
 # Ensure we have curl or wget support.
 #
 
-CURL_PARAMS=( "-L"
+CURL_PARAMS=( "-k"
+              "-L"
               "-#")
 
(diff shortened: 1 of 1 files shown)


qw3rtman:
I assume SSL is not required with the `wget` code so `n project` works nicely. Can anyone confirm?

bitdivine:
Should not the default be secure?  A given user might declare explicitly that they don't mind being pwned and do so by giving --insecure as a flag but most users should not have to type --secure to insist that certificates be checked.  Would you be OK with an --insecure flag being added to the n command line?  I can prepare a pull request.

norpol:
Yes, this merge request should be closed. Having insecure as default is a bad habit.

PR #345, opened 2016-03-06
replace bash with sh

### LINKED ISSUE #344: Use `sh` instead of `bash`
For use in minimal environments, such as [alpine](http://www.alpinelinux.org/) which does not have bash installed by default.

theRemix:
resolves #344

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 #
 # Setup.


troy0820:
This issue #344 is closed because of this comment 

> This patch is not sufficient. If one just changes bash to sh, then the script will fail:
/usr/bin/n: line 33: syntax error: unexpected "("

I'm going to close this PR because pulling this in would break the functionality.

PR #348, opened 2016-03-23
Handling SIGINT correctly.

### LINKED ISSUE #320: No input echo'ed after listing installed versions
- OSX 10.10.5
- Bash 4.3.42

Occurs in iTerm2 and OSX Terminal.

Steps to reproduce:

- `n`
- `CTRL-C`
- Entered keys and pasting from clipboard is not shown anymore

JeffDownie:
This PR seems to fix #320 on my system (Ubuntu 15.10) - the fix itself I took from [here](http://permalink.gmane.org/gmane.comp.shells.bash.help/1482) and [here](http://www.cons.org/cracauer/sigint.html).

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -134,7 +134,7 @@ leave_fullscreen() {
 
 handle_sigint() {
   leave_fullscreen
-  exit $?
+  kill -INT $$
 }
 
 handle_sigtstp() {


KevinSjoberg:
Would love to see this merged. Currently running `n` to output the versions followed by `C-c` will cause problems as described in #320.

qw3rtman:
Thanks, @JeffDownie!

qw3rtman:
Hmm, I should have checked before merging (which I've now reverted), but this change breaks `n` when I do `C-c` on OS X. It goes into an infinite loop of `bash`, `stty`, and `tput`.

Perhaps we'll need to handle this specifically for different OS's. What are you running, @KevinSjoberg?

KevinSjoberg:
I'm running OS X and I have not actually tried this patch more than that I acknowledge the problem. I'm up for playing around with this though to make it cross-os compatible.

JeffDownie:
@KevinSjoberg I didn't/don't have any OS X machines to test this against. Might be worth knowing what version of bash you are running, as it could be bash handles it differently between different versions, rather than the OS being to blame - for reference, I'm currently on 4.3/4.4.

PR #360, opened 2016-07-10
Some mirror site lts version error !

lonsenluo:
Some mirror site stored more then one lts version.So need sort to select max version.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -661,6 +661,7 @@ display_latest_lts_version() {
   $GET 2> /dev/null ${MIRROR[$DEFAULT]}/$folder_name/ \
     | egrep "</a>" \
     | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' \
+    | sort -u -r \
     | head -n1
 }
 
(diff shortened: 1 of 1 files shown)


qw3rtman:
I think the proper fix would be to [line 658](https://github.com/tj/n/blob/master/bin/n#L658).

Additionally, simply calling `sort -u -r` could create problems (it would say `1.2.15 < 1.2.4`, which is not true). Instead `sort -u -k 1,1n -k 2,2n -k 3,3n -t` should be used as done on [line 645](https://github.com/tj/n/blob/master/bin/n#L645) and [line 677](https://github.com/tj/n/blob/master/bin/n#L677).

Thanks for pointing this out!

JohnRGee:
Example mirror: https://npm.taobao.org/mirrors/node/latest-v8.x/

JohnRGee:
Also need to change the `head` to `tail` in the line following the sort to get the latest version. See #512 which has this change too.

JohnRGee:
I think the sort got changed from reverse order to normal in the extra commit, so the `head` needs to change to `tail` to match (and the other filters use tail). 

However, there is a new PR now, so no need to update this one and could just close.

Issue #361, opened 2016-07-12
N displaying incorrect version when entered with -V


paolord:
I installed N with `npm install -g n@2.1.2` since npm says it's the latest release, but after installation when I checked it with `n -V` it displays 2.1.0.

PR #362, opened 2016-07-12
updated VERSION variable to 2.1.1 to mirror version in package.json

paolord:
Posted an issue for this #361, entering n -V outputs a different version from package.json

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -4,7 +4,7 @@
 # Setup.
 #
 
-VERSION="2.1.0"
+VERSION="2.1.3"
 UP=$'\033[A'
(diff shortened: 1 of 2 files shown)


qw3rtman:
Thanks for the PR. I think we'll have to bump it to `2.1.3` though in order to account for this commit.

paolord:
You're welcome. I fixed my PR to have both the bash script and the package.json to have the same value as the next version, 2.1.3, hope it helps.

qw3rtman:
Perfect, thanks! 😄

PR #364, opened 2016-07-22
Persist global npm installs

Jonnymcc:
Sometimes I need a different version of npm than the one installed with a version. Every time I switch node versions with n it wipes out my global node modules since it is a fresh copy from that installed version. This change backs up the global node modules so that the next time you switch to a previous version you were using the global node modules you installed will still be there.

run `./run_tests.sh` to demonstrate that this change fixes the problem.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -487,14 +487,14 @@ activate() {
   if test "$version" != "$active"; then
     local old_dir=$BASE_VERSIONS_DIR/$active
     local dir=$BASE_VERSIONS_DIR/$version
+    if test -d "$N_PREFIX/lib/node_modules"; then
+      cp -fR "$N_PREFIX/lib/node_modules" "$old_dir/lib"
+    fi
     # Remove old npm to avoid potential issues with simple overwrite.
     if test -d "$dir/lib/node_modules/npm"; then
       if test -d "$N_PREFIX/lib/node_modules/npm"; then
         rm -rf "$N_PREFIX/lib/node_modules/npm"
       fi
-    fi
-    if test -d "$N_PREFIX/lib/node_modules"; then
-      cp -fR "$N_PREFIX/lib/node_modules" "$old_dir/lib"
     fi
     # Copy (lib before bin to avoid error messages on Darwin when cp over dangling link)
     for subdir in lib bin include share; do
@@ -801,3 +801,4 @@ else
     shift
   done
 fi
+
--- a/bin/n
+++ b/bin/n
@@ -485,12 +485,16 @@ activate() {
   local version=$1
   check_current_version
   if test "$version" != "$active"; then
+    local old_dir=$BASE_VERSIONS_DIR/$active
     local dir=$BASE_VERSIONS_DIR/$version
     # Remove old npm to avoid potential issues with simple overwrite.
     if test -d "$dir/lib/node_modules/npm"; then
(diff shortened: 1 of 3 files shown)


qw3rtman:
I think this is a great addition; however, it is extremely slow to copy every single global module installed. For this reason, I think persisting a global module should be opt-in. For example:

```sh
n 0.11.1 # install v0.11.1
npm install --global yo@1.1.0
yo --version # 1.1.0
n persist yo # persist the current `yo` version for this version of Node

n latest # install v6.3.1
npm update --global yo
yo --version # 1.8.4
n persist yo # persist the current `yo` version for this version of Node

n 0.11.1 # switch back to v0.11.1
yo --version # 1.1.0
```

There could be a similar command to unpersist.

What do you think about this?

Jonnymcc:
I think making it opt in is fine. I feel that adding a command to persist once or to toggle is a bit much. Instead I'm thinking setting an environment variable that if set persists global modules is a good solution. For people that want to persist their global installs I believe that they would want this to apply to all versions managed by `n` and not have to worrying about running a command. In most cases I would imagine that globally installed modules would be light, I'm thinking npm, grunt, bower?.. and that the bulk of installed modules would be installed locally to a project.

Does adding an environment variable sound sufficient?

JamesChevalier:
I just switched from homebrew's node to having `n` manage my node installs. So, I'm brand new to `n` & my background is mostly in ruby. This probably skews my opinions quite a bit, so my perspective might not fall in line with the node community.

Just in case it's helpful... I expected that each version of node that `n` installed would have its own "global" view of npm & its packages.
I thought that `n 6.6.0 && npm -g list` and `n 4.5.0 && npm -g list` would produce two different lists of packages. I think my mindset here is coming from gemsets/bundler.

I don't know if this is a completely ridiculous suggestion. I just wanted to share my point of view as a relative newcomer, in case it's useful.

Thanks for all your hard work!

qw3rtman:
@JamesChevalier I think isolating `npm` for each Node version is important. At the same time, I think it should be implemented to accommodate `n`'s ease-of-use. There are some users of `n` (myself included) who don't want to manually manage packages for each Node version. There are also some users who would benefit from that kind of flexibility.

As such, my main concern is that a solution compensating for both user's needs would be

1. too automatic (makes many assumptions) for *power users*, or
2. too error-prone (configuration can become a mess) for more *relaxed users*.

If we are able to come up with a solution that serves both types of users, I would be all for implementing it! If not, making this optional may be best.

Thanks for sharing your thoughts! Community input is what makes good open-source software great. 😄

Kikobeats:
what about that?

sam3d:
What about what, @Kikobeats?

shadowspawn:
This is interesting, but I do not think adding more files to the install is a good fit for `n`. See especially this comment: https://github.com/tj/n/pull/364#issuecomment-250897760

I think a lighter weight approach to achieve the separation would be to symlink into nvh/versions from the install destination instead of copying, but that isn't something I plan to do either.

Feel free to open a new issue if this comes up again, with new information and renewed interest.
 
shadowspawn:
I have added support for preserving npm in [nvh](https://github.com/shadowspawn/nvh) if any readers would like to give it a try. (It was made possible by switching to rsync for the install and so can't be brought back to n.)

PR #369, opened 2016-08-04
Add older Mac support to custom architecture section of the Readme

StewartNoyce:
In response to @qw3rtman, I looked more closely at the potential for a pull request.  From the n perspective, you can actually load LTS (4.4.7) onto the older Intel Core 2 Duo systems (like my now ancient Mac Mini) with the custom architecture function.  It worked for me, and node 4.4.7 ran ok on the command line.  I updated the Readme file to reflect that.  Now, what people do with it is up to them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,10 @@ Download and use latest 32 bit version of `iojs`:
 
     $ n io --arch x86 latest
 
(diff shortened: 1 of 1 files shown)


qw3rtman:
Looks great. Thanks!

PR #371, opened 2016-09-08
Resolve versions with no dots

MiniGod:
Before:

```
$ node -v
v4.5.0
$ n 6

     install : node-v6

  Error: invalid version 6

$ node -v
v4.5.0
```

After:

```
$ node -v
v4.5.0
$ n 6

     install : node-v6.5.0
       mkdir : /home/user/n/n/versions/node/6.5.0
       fetch : https://nodejs.org/dist/v6.5.0/node-v6.5.0-linux-x64.tar.gz
   installed : v6.5.0

$ node -v
v6.5.0
```

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -500,7 +500,7 @@ install() {
   local version=${1#v}
 
   local dots=$(echo $version | sed 's/[^.]*//g')
-  if test ${#dots} -eq 1; then
+  if test ${#dots} -lt 2; then
     version=$($GET 2> /dev/null ${MIRROR[DEFAULT]} \
       | egrep "</a>" \
       | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' \


qw3rtman:
One issue this could lead to, as seen in the example provided, is installing a version of Node the user wasn't expecting. Typing `n 6`, a user may expect to install `Node v6.0.0` (i.e: `n 6` is shorthand for `n 6.0.0`).

I think the behavior should be as follows.

```sh
n 6 # n 6.0.0
n 6.0 # n 6.0.0
n 6.5 # n 6.5.0
```

Perhaps if this checks if, after adding the required `.0`'s, the versions are the same, it would be more predictable.

MiniGod:
I think the behavior is good. Compare it to installing a module with npm.
If i run `npm install -g npm@2`, I expect to get the latest version of npm v2, and not `v2.0.0`

qw3rtman:
Ah, I see. This behavior does make more sense.

qw3rtman:
Thanks for your contribution! 😄

PR #396, opened 2016-11-08
Makes n compatible with ash and its family

### LINKED ISSUE #395: Error running on OpenWRT
N requires bash to run. I linked the ash shell, used by OpenWRT, to /bin/bash. When I run `# n latest`, the output is `bash: applet not found`. Is there a way to run n without calling bash? Or work around this somehow?

sdaitzman:
Tested on my MacBook

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 #
 # Setup.


sdaitzman:
Does not work on LinkIt. I'll continue testing and try to find a solution.

qw3rtman:
On macOS, `sh -> bash`; however, this may not always be the case (see [this](http://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang)). For this reason, it's best to use `#!/usr/bin/env bash` for portability across machines that do not link `sh` to `bash`.

Issue #404, opened 2017-01-06
n fails when there is no internet access and other semi-related issues


sine9:
Hello n developers! I love `n` and use it to quickly switch versions of node for testing.

I was on a plane recently no internet access and n stopped working.  After seeing that you needed to specify the exact version number, I attempted to write a shell script wrapper and found other issues worth noting.  They are all listed here:

1) there seems to be no way to tell `n` to use the latest available for a given semver description if internet access is not available.  To test this, try turning off your internet and running

```
$ sudo n 7
$ sudo n 7.
$ sudo n 7.x
```

All of these fail.  The only thing that worked was running `n` and pulling the exact version number.

2) a version number with an asterisk fails with some cp errors:

```
$ sudo n 7.*.*
cp: /usr/local/n/versions/node/7.*.*/bin: No such file or directory
cp: /usr/local/n/versions/node/7.*.*/lib: No such file or directory
cp: /usr/local/n/versions/node/7.*.*/include: No such file or directory
cp: /usr/local/n/versions/node/7.*.*/share: No such file or directory
```

`n` probably should just fail rather than attempt a command.

3) `n ls` gives blank input when there is no internet connection.  I can understand why, since "available" is determined by comparing against a list that is unavailable when internet is off, but since `n` can switch between installed versions `n ls` should show only the installed versions when internet access is not available.

4) There is no command to just get a plaintext list of the installed versions.  `n` is an interactive tool, redirecting to file or pipe screws with the shell session.  `n ls` shows in bold but it's difficult to work with that programmatically.  Some TTY detection logic may be necessary.

5) the `n` list sorts the `0.10` and `0.12` series before the `0.8` series. That seems to be a bug that will rear its ugly head again if `6.10` is reached.

6) How do I remove all older minor or patch versions of node?  For example, my computer has a bunch of older node versions and I would like to prune to the latest patch version of the given major version:

```
    node/0.8.28

    node/0.9.12

    node/0.10.46 <-- remove
    node/0.10.47 <-- remove
    node/0.10.48

    node/0.11.16

    node/0.12.15 <-- remove
    node/0.12.16 <-- remove
    node/0.12.17 <-- remove
    node/0.12.18

    node/4.5.0 <-- remove
    node/4.6.0 <-- remove
    node/4.6.2 <-- remove
    node/4.7.0 <-- remove
    node/4.7.1

    node/5.12.0

    node/6.1.0 <-- remove
    node/6.8.0 <-- remove
    node/6.9.1 <-- remove
[...]

PR #407, opened 2017-01-24
Add pruning for non-active versions

sam3d:
Add a basic prune command to remove all versions that aren't currently active.

```
n prune                        Remove all versions except the current version
```

Current output of running `n prune` with only `7.1.0`, `7.2.0` & `7.4.0` installed (with `7.4.0` being the active version):

```
$ n prune
node/7.1.0
node/7.2.0
```

Haven't extensively tested it yet but it seems to work well! Any feedback about output displayed back to the user or command syntax would be appreciated, but otherwise would love to get this implemented.

Kind of refs #404, but is not as feature complete as the suggestion. Additional functionality to the prune command could be easily added with CLI flags, such as `n prune --minor` or `n prune --all`.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -167,6 +167,7 @@ display_help() {
     n use <version> [args ...]     Execute node <version> with [args ...]
     n bin <version>                Output bin path for <version>
     n rm <version ...>             Remove the given version(s)
+    n prune                        Remove all versions except the current version
     n --latest                     Output the latest node version available
     n --stable                     Output the latest stable node version available
     n --lts                        Output the latest LTS node version available
@@ -578,6 +579,14 @@ remove_versions() {
 }
 
 #
+# Prune non-active versions
+#
+
+prune() {
+    abort "not implemented"
+}
+
+#
 # Output bin path for <version>
 #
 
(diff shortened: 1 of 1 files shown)


sam3d:
Any feedback @qw3rtman or @troy0820?

troy0820:
LGTM but I want to see if @qw3rtman is okay with pulling in the changes.  Right now we are at 2.1.4 and I'm not sure where this fits in the release if and when we bump the version. 👍

sam3d:
Sorry don't mean to be pushy; it's been a fortnight since making the PR and just wanna bump to see if @qw3rtman has any feedback 😄

sam3d:
Does `n` need a new top-level maintainer? I mean this in the most friendly way possible :) @troy0820 you seem perfectly suited for the role, as @qw3rtman just doesn't appear to engage (though this is not a criticism). It seems only fair you should be able to make decisions unilaterally given this.

troy0820:
I would have to be given top level access but that is at the discretion of the other maintainers here in this repository.  I would do my best merging PRs and triaging issues.

troy0820:
Closing because #422 was merged 👍

Issue #401, opened 2016-12-28
Doesn't render with solarized terminal color scheme


owais:
What it looks like:

<img width="670" alt="screen shot 2016-12-29 at 1 36 31 am" src="https://cloud.githubusercontent.com/assets/46186/21530514/63c1af54-cd67-11e6-8265-f47c12cb9460.png">

-------------

What it looks like after selecting text:

<img width="659" alt="screen shot 2016-12-29 at 1 36 37 am" src="https://cloud.githubusercontent.com/assets/46186/21530523/6d36df00-cd67-11e6-87f4-3d64a07192f2.png">

-------------------

I'm using iTerm2 Build 3.0.13 on OS X with Fish Shell and Solarized color scheme.

sercanov:
+1, dark fonts are not visible

troy0820:
Issue resolved with PR #412

PR #412, opened 2017-02-09
using DIM instead of Dark gray when listing and logging

cocoanton:
Using DIM instead of darkgray color to improve readability on solarized color themes.
solarized
solves issue #401 

Tested with different color themes in iTerm2 to see that it works. Might need more testing for other terminal programs.
different-themes

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -15,7 +15,7 @@ BASE_VERSIONS_DIR=$N_PREFIX/n/versions
 #
 
 log() {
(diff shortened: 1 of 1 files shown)


troy0820:
NICE!!!  I'm going to pull this down and run this in a container to test other OS. Plus I would like to get some feedback from the other contributors before I pull it in and bump the version.

troy0820:
Tested it in a container and it LGTM.  👍 Thank you for your contribution @cocoanton

Issue #304, opened 2015-08-30
Handling SIGINT on OS X doesn't show expected behaviour


maxrimue:
When exiting `n` while it's in fullscreen (with ctrl+c), `leave_fullscreen()` isn't executed and the shell remains weird until `tput rmcup` and `stty echo` are executed manually.

courajs:
I came here to report this issue exactly

mashiox:
This occurs on Ubuntu 16, too.

troy0820:
I can't reproduce the problem to help with this issue.  I have a docker image `troy0820/node-n:16.04` that is an Ubuntu 16 and I can't reproduce it.  What version of `n` are you running?

mashiox:
@troy0820 Whatever's on npm. I'm making a PR with my fix right this moment

PR #424, opened 2017-03-30
Kill n's process group on sigint

mashiox:
# Pull Request Template:

### Describe what you did
 _On sigint, kills n's process group, rather than only exiting with the parent's exit status_

### How you did it
Replaced `kill $?` with `kill 0`, which will [kill a process group](http://unix.stackexchange.com/a/67552/192793)
__

It does effect n's functionality. Previously, bash would hang with the process group, and the terminal might not echo the input from STDIN. This kills all of those hanging processes and gives the user STDIN back.

This addresses issue #304

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -4,7 +4,7 @@
 # Setup.
 #
 
-VERSION="2.1.6"
+VERSION="2.1.7"
 UP=$'\033[A'
 DOWN=$'\033[B'
 N_PREFIX=${N_PREFIX-/usr/local}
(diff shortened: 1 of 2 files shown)


troy0820:
Do me a favor @mashiox, can you change the version in the package.json to `2.1.7` and in `bin/n` as well?  I'll merge it in after that.

mashiox:
@troy0820 aye, aye! Pleasure working with you!

troy0820:
Definitely, appreciate your PR to N.  Thanks a lot @mashiox 👍

PR #426, opened 2017-04-06
Ignore broken certificates when using n over ssl proxy

alqu:
In wget there is already the "--no-check-certificate" option, but in curl it was missing. Added.

# Pull Request Template:

### Describe what you did
I added  an additional param for curl that ignores invalid ssl certificates. Otherwise, the download fails and leaves a broken install behind.

### How you did it
I tried to pull a newer version of node with `n latest`  through an ssl proxy. Curl was complaining that the certificate was not valid - which is true for SSL proxies a.k.a. MITM.

Fetching updates with a "normal" environment works flawlessly.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -54,6 +54,7 @@ fi
 #
 
 CURL_PARAMS=( "-L"
+              "--insecure"
               "-#")
 
 WGET_PARAMS=( "--no-check-certificate"


dskrvk:
This shouldn't be enabled by default in either place (#475 removes the option for wget). If some users absolutely need this and are OK with sacrificing security, there can be an option to skip certificate checks.

dskrvk:
Actually, the same change has already been submitted and received a similar comment in #324.

shadowspawn:
Proxy setup is a bit tricky @alqu, but as noted by @dskrvk the default for `curl` and `wget` should be secure.

alqu:
Thanks!

Meanwhile, this feature request is not that urgent nor valid anymore, so I am fine with it.

PR #442, opened 2017-06-08
Support Cygwin & MSYS under Windows

baip:
# Pull Request Template:

### Describe what you did
Support Cygwin & MSYS under Windows

### How you did it
Account for the differences in directory structure and the file type of the installer package.

The behavior changes only if `uname -a` outputs CYGWIN or MSYS.

N/A

N/A

Done

Support Cygwin & MSYS under Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Node.js version management: no subshells, no profile setup, no convoluted API, j
 
 ![](http://nimit.io/images/n/n.gif)
 
-(Unfortunately `n` is not supported on Windows yet. If you're able to make it work, send in a pull request!)
+(`n` needs a bash shell from [Cygwin](https://www.cygwin.com) or [MSYS2](http://msys2.org)|[Git for Windows SDK](https://git-for-windows.github.io) to work on Windows.)
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


troy0820:
I'm gonna try this code in the environments we support and then release a new version.  This may take some time but I am still looking at this.

davidreghay:
Does this imply support for Git Bash on Windows as well?

baip:
@troy0820 Hi, just wondering if you had a chance to try this out and if there is anything else I can do. Thanks!

@davidreghay Yes, Git Bash on Windows basically wraps MSYS under the cover.

shadowspawn:
I do not have easy access to Windows, so a question in meantime (I wish there was a Windows docker container!).

Does installing node with these changes only update node and npm for Cygwin & MSYS command line, or does it affect cmd and PowerShell too?

baip:
Since WSL came out, I no longer use Cygwin or MSYS. But I thought `n` is a bash program -- how is it supposed to run under CMD or PowerShell?

shadowspawn:
I meant if you change the version of node using `n` under Cygwin, would it affect the version of node and npm for cmd? (I am not sure how much they share search paths.)

baip:
For my old setup, I would install node using `n` under Cygwin and then add the directory to system `PATH`, so yes, in this case, both CMD and Cygwin use the same node installation and changes are shared. But if one doesn't manually add the directory to Windows `PATH` environment variable, then CMD knows nothing about it.

shadowspawn:
Thanks for info @baip 

With WSL available, I am not sure there is enough interest in an `n` running under Cygwin and MSYS  when there are Windows native node version managers available for non-WSL.

I am closing this Pull Request for now, but available for reference if this gets popular.

Thank you for your contributions.

Issue #411, opened 2017-02-01
vim-like shortcuts to select versions


rodrigomaia17:
it would be awesome if we could use J and K keys to select version as an alternative to the traditional arrow keys.

troy0820:
I can't find the keys with the escape sequence that's necessary for this feature.  I'm not sure if this is possible.

troy0820:
Issue resolved by PR #446

PR #446, opened 2017-06-29
Adds vim-like keybindings to version selection

zachfedor:
# Pull Request Template:

### Describe what you did

- modifies switch statement to differentiate escape sequences from
  single key presses
- watches `j` and up arrow to move selection up
- watches `k` and down arrow to move selection down
- watches `q` to quit selection
- watches for enter to make selection
- disregards other keys

I didn't change the version number yet. Not sure if you wanted a minor version update or wait to merge for a major release.

### How you did it

Instead of using `read` to watch for three characters at a time, I'm watching for one character. Given a single key, like `j` or `k`, we can use a simple string comparison to determine the function to run. If that key is an ANSI escape code, we know that other characters will be incoming. For example, hitting the up arrow in a terminal sends `\033`, `[`, and `A`. So we can use another `read` for a single character with a timeout to cancel if nothing follows the escape. If this next character is a `[`, then it might be an arrow key. Use another single timed `read` to see which arrow it is and call the correct function.

Run `n` and use arrow keys and `j`/`k` to test movement. Hit enter to select or `q` to quit. `Ctrl-C` still sends an interrupt as normal, but all other keys are disregarded.

Solves #411 as well as unreported issue with `n` hanging after one non-arrow key press because `read` was waiting for two more characters before continuing the script.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -5,8 +5,6 @@
 #
 
 VERSION="2.1.7"
-UP=$'\033[A'
-DOWN=$'\033[B'
 N_PREFIX=${N_PREFIX-/usr/local}
 BASE_VERSIONS_DIR=$N_PREFIX/n/versions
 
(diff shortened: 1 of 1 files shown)


troy0820:
I'm gonna check this out and merge it in hopefully soon.

troy0820:
Tested with Ubuntu:16.04 docker image 👍 Great work @zachfedor. Definitely, appreciate it.

PR #461, opened 2017-09-13
Add `lsi` and `installed` options to list installed versions

blakek:
### Describe what you did
Add new options to list installed node/iojs versions:

`n lsi`
`n installed`

**Questions**

- Should these have different names?
- Alternatively, should this instead be a flag on the `ls` option?
  * I was going to add this as a flag to `ls` originally, but `ls` only lists node or iojs where this should probably list everything installed.

### How you did it

Function already existed.  It was just added it to the options loop.

The function was already there.  The only change was to the options.  It would just keep users from installing versions named "installed" or "lsi".

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,10 +158,11 @@ Output can also be obtained from `n --help`.
 
     Aliases:
 
-      which   bin
-      use     as
-      list    ls
-      -       rm
+      which      bin
+      use        as
+      list       ls
+      installed  lsi
+      -          rm
 
 ## Custom source
 
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Simply execute `n <version>` to install a version of `node`. If `<version` has a
     $ n 0.9.6
 
 Execute `n` on its own to view your currently installed versions. Use the up and down arrow keys to navigate and press enter or the right arrow key to select. Use ^C (control + C) to exit the selection screen.
-If you like vim key bindings during the selection of node versions, you can use `j` and `k` to navigate up or down without using arrows. 
+If you like vim key bindings during the selection of node versions, you can use `j` and `k` to navigate up or down without using arrows.
 
     $ n
 
@@ -80,7 +80,7 @@ Alternatively, you can use `-` in lieu of `rm`:
 Removing all versions except the current version:
 
 ```bash
-$ n prune 
+$ n prune
 ```
 
 ### Binary Usage
@@ -135,6 +135,7 @@ Output can also be obtained from `n --help`.
       n --stable                     Output the latest stable node version available
       n --lts                        Output the latest LTS node version available
       n ls                           Output the versions of node available
+      n lsi                          Output installed versions of node and iojs
 
     (iojs):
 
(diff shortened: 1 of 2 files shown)


NathanJang:
```bash
$ n lsi
node/6.11.0
node/8.9.1
node/9.1.0
```

Right now, it lists versions with `node/`, etc. prefixed in front of them. Would it be better to print them in colour with pretty formatting, like how `n ls` does? We could copy the function `display_remote_versions`.

blakek:
I thought about using something similar to `display_remote_versions`.  The reason I didn't was:

- I mostly wanted something that was easily piped to other scripts.  However, thinking about it now, it would be extremely easy to test if running in a tty, so this doesn't matter.
- Everything listed is installed. I didn't see a reason to prefix the output with the colored circles since everything would have a circle before it.  However, I'd be happy to style it however and just switch to unformatted if not running in a tty.  What are your thoughts on how it should look?
- I don't like the `node/` prefix, but it differentiates node from iojs.  Maybe we could remove the `node/` prefix and show a prefix for things that aren't node proper??

NathanJang:
I think it's up to @tj, but I'd personally like this to follow the behaviour of `display_remote_versions`.
- `display_remote_versions` is pretty-formatted and doesn't look like it's intended to be piped, so I don't think this should either.
- There's only a circle next to the version currently activated, so only one version would have a circle next to it.

Thanks for your contribution! 😊

shadowspawn:
For comparison, `nvm` uses `ls` for the local versions and `ls-remote` for the remote versions.

Is there other software which uses `installed` in this way? My first tries were HomeBrew and npm but neither of them do.

In the bigger picture, I am concerned about ambiguity whether "installed" refers to cached (e.g. `/usr/local/nvh/versions/node/v10.14.1/`) or to copied from the download cache and made active (e.g. `/usr/local/bin/node`).

blakek:
@shadowspawn Most things I know of use `list`, `ls` or something similar to list local installs and something like `search` to look for a remote installation. `list` was already taken, so I just thought of another name. Would something like `list-local` be better?

@NathanJang I just updated to the behavior you mentioned:

```
$  ./bin/n lsi
```

example output:

```
    10.15.3
  ο 11.12.0
```

shadowspawn:
For comparison, both `nvs` and `nvm` have `ls` and `ls-remote`.

shadowspawn:
blakek Good observation that can detect tty, so pretty does not cause problems for scripts. Otherwise I would have suggested should be plain text as you did first time, since can use `n` and menu to see locally cached versions.

shadowspawn:
1) The `node/` prefix is because in theory there could be `iojs` installed or a custom project using `PROJECT_NAME`. So not appropriate to delete it while they are supported.

2) `lsl | list-local` seems the most correct name for now (although not as catchy as `installed`). Any naysayers?

3) I don't want another pretty printing routine. What about simply calling into the code used by the menu itself? e.g.

```
      lsl|list-local) check_current_version; display_versions_with_selected $active; exit ;;
```

If all of that is compelling enough, I'll close this Pull Request and consider a new one against the `develop` branch.

shadowspawn:
No further activity in a month. Closing as stale.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

Issue #335, opened 2015-12-29
n stable installs the version before stable.


callahanp:
https://nodejs.org/en/ shows v5.3.0 as stable  n stable downloads v5.2.0
# n stable

   install : node-v5.2.0
       mkdir : /usr/local/n/versions/node/5.2.0
       fetch : https://nodejs.org/dist/v5.2.0/node-v5.2.0-linux-x64.tar.gz
   installed : v5.2.0

![screenshot from 2015-12-29 13-00-28](https://cloud.githubusercontent.com/assets/6132176/12039197/5e7dac56-ae2c-11e5-8008-fab671d855c6.png)

qw3rtman:
Hmm, I'm not having this issue. Which OS and `n` version are you running?
<img width="697" alt="screen shot 2015-12-29 at 12 12 30 pm" src="https://cloud.githubusercontent.com/assets/1139621/12039375/9054e0de-ae25-11e5-9a4c-78e20307fde5.png">

tedgaydos:
The stable keyword will only grab even number minor versions. It is still using the old way of finding node versions. To get the current version, use the latest keyword.

The stable keyword needs to be refactored to be an alias of lts.

JohnRGee:
I got confused by this too when node was at 8.7. I have submitted a pull request which minimises code changes, and downgrades stable to an alias for lts in the documentation.

troy0820:
![screen shot 2018-11-05 at 3 41 25 pm](https://user-images.githubusercontent.com/5985850/48025406-4ef25c80-e111-11e8-995e-aaaad7652ffe.png)
Now that LTS is considered stable this issue is closed with PR #467

Issue #354, opened 2016-04-27
"Stable" to "Current"


tleb:
> We will be using the `Current` label from here-on instead of `Stable`.
> — https://nodejs.org/en/blog/release/v6.0.0/#current-what-happened-to-stable

tedgaydos:
Thanks for the info. We will take it into consideration.

The dist folder was restructured. Looks like we can easily support commands like `n latest`, `n latest v0.10`, `n latest v4`, `n io latest v1`, and `n lts`.

We would also remove the `stable` keyword.

shadowspawn:
Fixed `stable` in #467
More recent and extensive comments in #522

This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #467, opened 2017-10-27
Effectively make stable an alias for lts. (Solves issue #335)

JohnRGee:
# Pull Request Template:

### Describe what you did

Made stable an alias for lts.

### How you did it

Removed stable from documented options, other than as an alias for lts.
Modified display_latest_stable_version to return LTS version.
Retained existing stable code for backwards compatibility. 

n --latest: 8.8.1
n --lts: 6.11.5
n --stable: 6.11.5
n latest && node --version: v8.8.1
n stable && node --version: v6.11.5
n latest && node --version: v8.8.1
n lts && node --version: v6.11.5

Solves issue #335 where stable was using old pattern of latest even minor version number.

Using lts is a good match for "stable", but is a significant change to the erroneous behaviour which was sometimes same as latest .

Say node latest was at 8.9.2.

Old:
n --latest: 8.9.2
n --lts: 6.11.5
n --stable: 8.8.1

New:
n --latest: 8.9.2
n --lts: 6.11.5
n --stable:  6.11.5

Change stable to be an alias for lts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Simply execute `n <version>` to install a version of `node`. If `<version>` has 
     $ n 0.9.6
 
 Execute `n` on its own to view your currently installed versions. Use the up and down arrow keys to navigate and press enter to select. Use `q` or ^C (control + C) to exit the selection screen.
-If you like vim key bindings during the selection of node versions, you can use `j` and `k` to navigate up or down without using arrows. 
+If you like vim key bindings during the selection of node versions, you can use `j` and `k` to navigate up or down without using arrows.
 
     $ n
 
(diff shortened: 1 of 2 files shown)


JohnRGee:
Issue #354 also mentions no longer using Stable.

JohnRGee:
A different approach to deprecation if preferred could be to remove `stable` from documentation, but retain existing behaviour. This has the potential advantage of changing no existing behaviour while still avoiding new users tripping over the old meaning of  the `stable` label.

New references:
New user using `n install stable` [sic] mentioned in #485
Suggested terminology: #522

JohnRGee:
Rebased to resolve merge conflicts.

(Sorry about the merge conflict, due to whitespace changes, didn't notice I have my editor set to trim trailing spaces.)

JohnRGee:
(If you wish, I will update Pull Request again to remove the two extraneous whitespace changes. Would have done it during the rebase if I had realised at the time.)

PR #475, opened 2017-11-17
Fix insecure downloading of node-binaries for wget

norpol:
♥

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -54,8 +54,7 @@ fi
 CURL_PARAMS=( "-L"
               "-#")
 
-WGET_PARAMS=( "--no-check-certificate"
-              "-q"
+WGET_PARAMS=( "-q"
               "-O-")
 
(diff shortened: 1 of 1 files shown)


norpol:
Thank you for bumping and linking to #426.

norpol:
@troy0820 Could you check this out maybe?

norpol:
Not interested in getting this merged anymore.
The only recommendation I would give anyone evaluating whether or not to use `n` is to checkout [nodejs.org](https://nodejs.org/en/download/package-manager/) suggested methods. For example if you need something that is solving similar problems as `n`, checkout [nvm](https://nodejs.org/en/download/package-manager/#nvm) or use your systems package manager (best solution from security aspect).

shadowspawn:
Prerelease available via npm i n@next for v6. Release notes: https://github.com/tj/n/releases/tag/6.0.0-0

PR #479, opened 2017-11-26
Fix OSS check

chinesedfan:
# Pull Request Template:

### Describe what you did
Fix function `is_oss_ok`.
  - the recursive call should use the whole rest part as an argument
  - the `awk` picks the wrong location with `\r` at the end

### How you did it
Fixed by,
  - wrap the rest by backticks
  - replace `awk` with `sed`

NODE_MIRROR=https://npm.taobao.org/mirrors/node/ n 8.9.0

- Solved #314.
- #318 tried to fix but failed, and #460 is not the best way.

- Before
```
[zhong@localhost n]$ NODE_MIRROR=https://npm.taobao.org/mirrors/node/ n 8.9.0

     install : node-v8.9.0

  Error: invalid version 8.9.0
```
- After (Very quickly in China! But the official nodejs.org is slow due to network censorship.)
```
[zhong@localhost n]$ NODE_MIRROR=https://npm.taobao.org/mirrors/node/ n 8.9.0

     install : node-v8.9.0
       mkdir : /usr/local/n/versions/node/8.9.0
       fetch : https://npm.taobao.org/mirrors/node/v8.9.0/node-v8.9.0-darwin-x64.tar.gz
   installed : v8.9.0
```

Only 1 commit.

Fix OSS check then able to download from special mirror sites.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -385,32 +385,20 @@ erase_line() {
 }
 
 #
-# Check if the HEAD response of <url> is 200.
-#
-is_ok() {
-  if command -v curl > /dev/null; then
-    $GET -Is $1 | head -n 1 | grep 200 > /dev/null
-  else
-    $GET -S --spider 2>&1 $1 | head -n 1 | grep 200 > /dev/null
-  fi
-}
-
-#
 # Check if the OSS(Object Storage Service) mirror is ok.
 #
 is_oss_ok() {
+  local headers=
   if command -v curl > /dev/null; then
-    if $GET -Is $1 | head -n 1 | grep 302 > /dev/null; then
-      is_oss_ok `$GET -Is $1 | grep Location | sed -Ee 's/Location: (.+\.tar\.gz).*$/\1/'`
-    else
-      $GET -Is $1 | head -n 1 | grep 200 > /dev/null
-    fi
+    headers=`$GET -Is $1`
   else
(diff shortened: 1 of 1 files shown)


chinesedfan:
Updated with an optimization that saving headers in a local variable. Now it only needs to send 2 requests to check OSS resources, while the old implementation needs 4 requests.

JohnRGee:
I agree `is_oss_ok` was broken, but I am unable to reproduce the failure as described. Is `n` still failing for you in the same way?

I tried reproducing with MacOS and centos:6.6 and ubuntu:12.04 and ubuntu:latest.

For your interest, I am trying a different approach by changing `is_ok`, no parsing, no extra requests:
```bash
function is_ok() {
  if command -v curl &> /dev/null; then
    ${GET} --fail --head "$1" &> /dev/null || return 1
  else
    ${GET} --spider "$1" &> /dev/null || return 1
  fi
}
```

JohnRGee:
Redirects:
- `CURL_PARAMS` includes `-L` so `curl` follows redirects
- `wget` follows up to 20 redirects by default according to [this page](https://www.gnu.org/software/wget/manual/html_node/HTTP-Options.html), see `--max-redirect`

(I checked a few Linux variations and Mac, but not exhaustive coverage.)

shadowspawn:
See #560 for open issue. Subscribe to that for updates and/or add 👍 to show interest

PR #485, opened 2018-01-10
add colons to reduce ambiguity

arjunpdhaliwal:
# Pull Request Template:

### Describe what you did

Reduced ambiguity in an error message. A friend was having trouble as he typed "n install stable" rather than "n stable". An error message with a colon would have immediately indicated what was wrong, but the current message made him think that the version was incorrect. Note that adding "install" is the syntax in nvm, so people transitioning might be confused by this.

example:

`Error: invalid version install`

vs

`Error: invalid version: install`

### How you did it

Added a colon to it.

The change is trivial, but can be verified by simply running the string with the abort function in a separate script.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -539,7 +539,7 @@ install() {
       | egrep ^$version \
       | tail -n1)
 
-    test $version || abort "invalid version ${1#v}"
+    test $version || abort "invalid version: ${1#v}"
   fi
 
   local dir=${VERSIONS_DIR[$DEFAULT]}/$version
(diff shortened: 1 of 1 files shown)


JohnRGee:
Recent issue making same mistake: #519

shadowspawn:
As this is a common (and reasonable) mistake I am considering just adding `install` as a functional command so no penalty for people using syntax from other utilities.

shadowspawn:
I have a couple of improvements in mind related to this confusion:
- add `install` to match common usage (npm, brew, nvm, nvs)
- consistently refer to parameters with single quotes `'parameter'`

I won't pull in this Pull Request. If I have not released an improvement within a couple of months, feel free to call me on it!

Thank you for your contributions.

PR #494, opened 2018-03-27
Detect arm64 architecture

jessaustin:
### Describe what you did
Added `arm64` architecture compatibility.

### How you did it
Added an entry for the "arm64" architecture to the architecture `uname` case statement in `tarball_url()`. Actually, the string to look for in the `uname` output is `aarch64`, but on the nodejs site that architecture is called "arm64".

```bash
$ uname -a
```
On any machine on which the output of `uname` does *not* contain the string "aarch64", this PR will have no effect.

I never filed an issue for this.

If your machine is on this architecture (e.g. a Samsung Chromebook Plus), it previously would have worked like this:
```bash
$ sudo n latest

     install : node-v9.9.0
       mkdir : /usr/local/n/versions/node/9.9.0
       fetch : https://nodejs.org/dist/v9.9.0/node-v9.9.0-linux-x86.tar.gz
/usr/local/bin/n: line 582: /usr/local/bin/node: cannot execute binary file: Exec format error
   installed : 
$ node --version
-su: /usr/local/bin/node: cannot execute binary file: Exec format error
```
After this change, it's more like this:
```bash
$ sudo n latest

     install : node-v9.9.0
       mkdir : /usr/local/n/versions/node/9.9.0
       fetch : https://nodejs.org/dist/v9.9.0/node-v9.9.0-linux-arm64.tar.gz
   installed : v9.9.0

$ node --version
v9.9.0
```

"add detection for **arm64**/**aarch64** architecture"

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -435,6 +435,7 @@ tarball_url() {
     *x86_64*) arch=x64 ;;
     *armv6l*) arch=armv6l ;;
     *armv7l*) arch=armv7l ;;
+    *aarch64*) arch=arm64 ;;
   esac
 
   if [ ${arch} = "armv6l" -a ${BIN_NAME[$DEFAULT]} = node ]; then


tlvince:
@troy0820 / other contributors; could this be merged? @krzysdz submitted the same PR and I was about to do the same until I found this. Thanks 👍

krzysdz:
I've just noticed that there is another pull request adding support for arm64 - #448 published on 11 Jul 2017 so almost 11 months ago...

jessaustin:
Haha yes #448 is better. It's 13 commits behind, but it works for both `arm64` and `aarch64`. Install with either:

    $ git clone https://github.com/joeledwards/n
    $ cd n
    $ make install
or

    $ npm i -g https://github.com/joeledwards/n

Issue #497, opened 2018-05-01
Wrong order with node/10.0.0


cedric25:
### Describe issue including what OS you are using
I'm using macOS High Sierra

### Describe what version of N you have
n --version output: 2.1.7

### Describe how you installed N
npm install -g n

### Steps to reproduce issue
1. `n latest` which would install node/10.0.0 at the time of writing
2. Install other versions, ex: `n 8.5.0` and `n 9.9.0`
3. List installed versions: `n`

### Describe the results you received
```
node/10.0.0
node/8.5.0
node/9.9.0
```

### Describe the results you expected
node/10.0.0 to be at the end:
```
node/8.5.0
node/9.9.0
node/10.0.0
```

### What version of npm you are using
npm --version output: 8.5.0

### Additional information you deem important

Sort is happening in `versions_paths()`.
I got the expected result with
`sort -k 1,1 -k 2,2n -t /`
instead of
`sort -k 1,1 -k 2,2n -k 3,3n -t .`
(Split on `/` instead of `.`)
But I haven't tested all possible use-cases...

troy0820:
Closing issue from merged in PR #498

PR #498, opened 2018-05-01
Fix order with node/10.0.0 tj/n#497

cedric25:
### Describe what you did
Change the sort logic when displaying installed versions. (Command `n`)

### How you did it
Example of a line: `node/8.5.0`
To do the sort, we were splitting the line on `.` characters, which would give [ 'node/8', '5', '0' ]
Then applying an alphabetical order on the first term, and a numeric order on the two next terms.
But it doesn't give the expected result when having `node/10.0.0`, as `node/10` is considered less than `node/8` when applying an alphabetical order.
I now split the line on `/` character, which gives [ 'node', '8.5.0' ], applying an alphabetical order on the first term and a numeric order on the second term.

### How to verify it doesn't affect the functionality of n
Try `n` command with a few different versions of io / node installed and check that the order is as expected.

solves issue #497 

Previously:
```
io/3.3.1
node/0.12.18
node/10.0.0
node/6.11.0
node/6.13.0
node/7.10.1
node/8.5.0
node/9.2.0
node/9.2.1
node/9.5.0
node/9.9.0
```

After fix:
```
io/3.3.1
node/0.12.18
node/6.11.0
node/6.13.0
node/7.10.1
node/8.5.0
node/9.2.0
node/9.2.1
node/9.5.0
node/9.9.0
node/10.0.0
```

Versions 10 or greater are now displayed as expected at the end of the list of installed versions

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -285,7 +285,7 @@ versions_paths() {
   find $BASE_VERSIONS_DIR -maxdepth 2 -type d \
     | sed 's|'$BASE_VERSIONS_DIR'/||g' \
     | egrep "/[0-9]+\.[0-9]+\.[0-9]+$" \
-    | sort -k 1,1 -k 2,2n -k 3,3n -t .
+    | sort -k 1,1 -k 2,2n -t /
 }
 
 #


troy0820:
Nice work on this PR 👍.  I'll merge and bump the version.

PR #499, opened 2018-05-03
Don't clear scrollback history in `display_versions`

ashtneoi:
Currently, n clears scrollback history in the `display_versions` function because it calls `clear`. (The fact that we're in fullscreen mode at the time apparently doesn't matter.) This PR replaces calls to `clear` with calls to `tput clear`, which only clears the visible screen, not the scrollback history.

The changes are trivial and `tput` is already known to exist by the time we run `tput clear` (because of the prior call to `tput smcup`), so this shouldn't affect existing functionality.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -322,7 +322,7 @@ list_versions_installed() {
 display_versions() {
   enter_fullscreen
   check_current_version
-  clear
+  tput clear
   display_versions_with_selected $active
 
   trap handle_sigint INT
@@ -342,26 +342,26 @@ display_versions() {
           read -rsn 1 -t 1 arrow
           case "$arrow" in
             $UP)
-              clear
+              tput clear
               display_versions_with_selected $(prev_version_installed)
               ;;
             $DOWN)
-              clear
+              tput clear
               display_versions_with_selected $(next_version_installed)
               ;;
           esac
         fi
         ;;
       "k")
-        clear
+        tput clear
         display_versions_with_selected $(prev_version_installed)
         ;;
       "j")
-        clear
+        tput clear
         display_versions_with_selected $(next_version_installed)
         ;;
       "q")
-        clear
+        tput clear
         leave_fullscreen
         exit
         ;;


JohnRGee:
What operating system has this behaviour? (e.g. On Mac, `clear` only clears the screen and not the scrollback history.)

JohnRGee:
I reproduced `clear` affecting scrollback in Ubuntu (from Mac using docker) but had same issue after changing to `tput clear`. I was intrigued and went digging.

Some links led me to the history section of [man page for clear](http://man7.org/linux/man-pages/man1/clear.1.html) where `tput clear` missed the E3 support for clearing scrollback in 2013 but caught up in 2016 and then shared logic with `clear`. However, both that page and the [man page for tput](http://man7.org/linux/man-pages/man1/tput.1.html) have a `-x` option to not attempt to clear the terminal scrollback, so a new approach possible.

ashtneoi:
@JohnRGee Sorry for being terse---I was on my phone. So it looks like my (Ubuntu 17.10) version of `clear` comes from ncurses, and the man page specifically says it clears the scrollback buffer.

JohnRGee:
(And to add to the possible scenarios, there was a bug with recognising `-x` with `tput clear` which was [fixed](http://invisible-island.net/ncurses/NEWS.html#index-t20180630) this year.)

ashtneoi:
Ugh, what a mess. I guess we could just echo the "clear screen" escape code manually, since any modern terminal will support it? unless I'm wrong and that's also platform-dependent.

JohnRGee:
There are a lot of other escape sequences in code, so maintainers may be fine with that approach.

JohnRGee:
I got annoying flicker when I tried simple escape code replacing calls to clear (despite it being apparently same as `clear` produces on my macOS).

```
function clear_screen() {
  # `clear`` and `tput clear` may clear scrollback, so just send simple clear screen ourselves.
  printf "\033[H\033[2J"
}
```

I got good visual results with using escape code as above for initial clear, and just setting cursor to home before redisplaying (and letting it overwrite).

```  
  # To avoid flicker with clearing screen, just reset cursor.
  PREPARE_FOR_REDISPLAY=$(tput home 2> /dev/null) || PREPARE_FOR_REDISPLAY=$(clear)
...
              printf "${PREPARE_FOR_REDISPLAY}"
              display_versions_with_selected $(prev_version_installed)
```

Surprisingly complicated!

JohnRGee:
The suggestion to avoid clear and use home when redrawing list is suboptimal when list is longer than screen. Rethinking...

JohnRGee:
Happier with second try.

```
  # Linux "clear" added clearing scrollback buffer as well as the screen. Try "clear -x" to opt out of that, if available.
  # (On Mac, "clear" is flicker-free compared with writing escape sequence ourselves, so use the actual command.)
  CLEAR="clear"
  clear -x &> /dev/null && CLEAR="clear -x"
```

JohnRGee:
I liked the `clear -x`approach until I found that Debian `clear` includes scroll-back, and ignores all parameters. I am giving up on robust solution! Good luck.

https://manpages.debian.org/stretch/ncurses-bin/clear.1.en.html

shadowspawn:
This was way harder than first appeared, and the proposed change would introduce issues in other configurations. 

This Pull Request has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 

Issue #502, opened 2018-05-09
Listing of installed versions is broken


kirisu:
# Issue Template:

### Describe issue including what OS you are using
macOS Sierra 10.12.6. 
#498 seems to have broken being able to list/switch between versions.

### Describe what version of N you have
2.1.10

### Describe how you installed N
N is installed via Homebrew.

### Steps to reproduce issue _1.2.3_
1. `n`

### Describe the results you received
```
sort: stray character in field spec: invalid field specification '2,2V' 

  Error: no installed version

  Usage: n [options/env] [COMMAND] [args]

...
```

### Describe the results you expected
A list of installed versions of node should be displayed, with the ability to switch. 8.11.1 is currently installed.

### What version of npm you are using
6.0.0

apertureless:
Got the same error

troy0820:
Can you do a `which node` and a `which n` because testing this in a docker container, it works.  I didn't install it through brew which I'm not sure why that would be a problem but the linking of certain things with brew complicate things as well.

apertureless:
```bash

❯ which sort
/usr/bin/sort


❯ sort --version
sort (GNU coreutils) 5.93
Copyright (C) 2005 Free Software Foundation, Inc.
This is free software.  You may redistribute copies of it under the terms of
the GNU General Public License <http://www.gnu.org/licenses/gpl.html>.
There is NO WARRANTY, to the extent permitted by law.

Written by Mike Haertel and Paul Eggert.
```

troy0820:
If you do a sort --help do you see the option for -V  ` -V, --version-sort          natural sort of (version) numbers within text`

Your sort version is older so I'm assuming they put the -V option in the newer version of sort.  The reason why we went to this is because when you list the versions with n, it is looking at the folders you have stored where node "lives" hence the `node/10.1.0` so when you select it it uses that directory for your "node".  The `n ls` is showing you what's available within the `https://node.org/dist/` and doesn't have the node/ before the version.

stsvilik:
Same here, had to remove `V` option from `2,2V` to at least list the versions

kirisu:
Can confirm that pointing to a newer version of sort patches the issue.

troy0820:
Can I close this issue? @kirisu @stsvilik @apertureless

stsvilik:
[...]

PR #504, opened 2018-05-13
Go back to plain numeric sorting for older versions of sort

JohnRGee:
# Pull Request Template:

### Describe what you did

Replaced introduced sort flags not supported in older versions of sort.

### How you did it

Used sed to make the prefix (node/) into a key (node.) so number parts can be sorted using (old) numeric sort, then restore prefix.

n listing now works on macOS Sierra 10.12.6, still works on macOS High Sierra 10.13.4.

Solves issue #502

Fix n interactive list to still work with older versions of sort (and still put node 10 after node 9).


troy0820:
Thanks @JohnRGee 👍

PR #509, opened 2018-06-07
Don't skip cert check when using wget

andreaso:
Curl checks TLS cert validity. Wget ought to do the same.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -54,8 +54,7 @@ fi
 CURL_PARAMS=( "-L"
               "-#")
 
-WGET_PARAMS=( "--no-check-certificate"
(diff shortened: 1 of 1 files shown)


JohnRGee:
I think this is appropriate both for security and for parity with curl implementation.

(If maintainers are worried about existing users of `wget` with broken certificates, would support for a new option like `--insecure` get this over the line?)

andreaso:
Pull requested timed out.

shadowspawn:
Prerelease available via npm i n@next for v6. Release notes: https://github.com/tj/n/releases/tag/6.0.0-0

PR #512, opened 2018-06-26
Fix n lts incorrect version

wmzy:
I could not install latest LTS version with `n lts` command:
```bash
$ n --lts
8.10.0
```
With this patch:
```bash
$ n --lts
8.11.3
```
I just sort and tail like `display_latest_version` and `display_latest_stable_version`.


JohnRGee:
Are you using a custom node mirror?

(I think this fix for the case when the mirror has more than one version in its "latest" folder, which is not normally the vase for the default mirror. Older similar PR in #360 )

wmzy:
Yes, I'm using the taobao mirror.

shadowspawn:
Merged to develop for next release.

Issue #488, opened 2018-02-07
pin npm version when switching node versions


reviewher:
`npm` 5.x have been causing issues with various modules, so locally I force `npm` 4.3.0.  Every time I switch to node 8 or 9 it installs some `npm` 5 version.  Is there a way to switch node versions with `n` without touching the current `npm` version?

JohnRGee:
`n` installs the version of `npm` bundled with `node`. There is not current a way to disable this.

(Discussed in #210.)

JohnRGee:
Implementation speculation: excluding npm might be reasonably straightforward if the install was done using rsync (using --exclude) rather than cp, but not sure about how compatible that would be.

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

shadowspawn:
I have added support for preserving npm in [nvh](https://github.com/shadowspawn/nvh) if any readers would like to give it a try. (It was made possible by switching to rsync for the install and so can't be brought back to n.)

shadowspawn:
`--preserve` added in `n` v6.3.0 to preserve the version of npm when installing node.

Issue #210, opened 2014-12-22
Switching node version changes installed npm


jarrodldavis:
Not sure why, but whenever I change the node version using `n`, my global `npm` is reverted to an old version and I have to run `npm install npm@latest -g` every time I switch. Is there a way I can tell `n` to ignore whatever internal `npm` it is using?

dstroot:
Actually the issue as I see it is that it seems impossible to keep npm updated properly - the npm version is "locked" to the node version it was installed with!  

For example if I install `n stable` I have npm version 1.4.28.  I then update npm to 2.1.7.  Then I switch to a different version of node and npm switches also.  OK I get that... but if I switch back to `n stable` I am back at 1.4.28 again.

There must be some way to update npm and have it stick, right?  Otherwise I can't use this amazingly handy utility.  :(

dstroot:
node and npm seem to be going their separate ways.  Here's how to decouple completely.
Now you can switch node versions with n all day and npm will remain stable as well as your global 
modules.  Cheers:

```sh
# Make a directory for global installations:
mkdir ~/.npm-global

# Configure npm to use the new directory path:
npm config set prefix '~/.npm-global'
# NOTE: add this to your profile as well:
export PATH=~/.npm-global/bin:$PATH 
npm install npm -g
echo "NPM global directory changed: $(npm config get prefix)"

# Remove old npm
sudo rm /usr/local/bin/npm
sudo rm -rf /usr/local/lib/node_modules
```

webdesserts:
Perhaps with the changes suggested in #212. Now would also be a good time to introduce a similar command such as `n npm 2.1.7` as well?

Globegitter:
Yep it would be great to be able to switch npm separately. Most of the time having to do `npm i -g npm` seems a bit cumbersome.

maxrimue:
I think all we have to do is to tell n to only copy `/lib/node_modules/npm` from the downloaded Node.js package if it's not already installed, right?

bucaran:
What do you think about [npmbrew](https://github.com/watilde/npmbrew)?

Globegitter:
Oh yeah, at a first glance npmbrew does look pretty good.

tjwebb:
> Actually the issue as I see it is that it seems impossible to keep npm updated properly - the npm version is "locked" to the node version it was installed with!

[...]

Issue #100, opened 2013-02-16
Error when copying `.../share/man`


hangtwenty:
Using Manjaro Linux 0.83 (XFCE Edition). Installed node 0.8.20 manually then used npm to install n. Have since used n to install several versions of node. Everything seems to be working, including switching node versions (as long as I do `sudo` but I figure that's given). I do get this error however:

```
cp: cannot overwrite non-directory ‘/usr/local/share/man’ with directory ‘/usr/local/n/versions/0.9.9/share/man’
```

Indeed, this is because `/usr/local/share/man` is a symbolic link to `/usr/local/man`. Not sure if this is applicable to Arch in general or just Manjaro.

Again, things still seem to be working; all that seems to be faltering is the copying of the man page, and that's fine really. Nonetheless thought it might be worth reporting.

qw3rtman:
Is this still an issue?

ntrrgc:
@qw3rtman Still happening in 2.1.3

qw3rtman:
@ntrrgc, what OS are you running?

ntrrgc:
Arch Linux.

deepsweet:
same error here with Arch Linux.

pcstl:
Same error happening on Antergos Linux. I'm copying manpages manually from ```/usr/local/n/versions/node/{version}/share/man``` to ```/usr/share/man``` as a workaround.

winneon:
Still happening. Has there been any progress on this?

rku4er:
I have noticed it works well if I run with sudo (Arch Linux)

devzsolt:
Still broken on Manjaro Linux. Any fix perhaps?

JohnRGee:
A manual work-around might be to install to a different location using `N_PREFIX`, and add to `PATH` or add symbolic links in `/usr/local`. Likewise, there is third party installer [n-install](
https://github.com/mklement0/n-install).

(A little history. The suggestion to fix this issue by using `rsync` instead of `cp` got rejected in #104. Support got added later to cope with symbolic links as direct subdirectories of `/usr/local` such as `/usr/local/bin` in #227, but not deeper in the hierarchy such as `/usr/local/share/man`.)

devzsolt:
[...]

PR #513, opened 2018-06-30
Add --preserve-npm

JohnRGee:
# Pull Request Template:

### Describe what you did

Added option for preserving npm during node install/activate.

### How you did it

Added a new flag and only changed behaviour if flag used. Used tar for the heavy lifting.

Code inspection that this adds a new code path, so existing functionality unaffected.

#210
#488

```sh
$ n 8.2.0
$ npm --version
5.3.0
$ npm install -g npm@6.0.0
...
$ npm --version
6.0.0
$ n 8.1.0
$ npm --version
5.0.3
$ npm install -g npm@6.0.0
...
$ npm --version
6.0.0
$ n -p 8.0.0
$ npm --version
6.0.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,10 @@ Flags also work here:
 
 ### Working with `npm`
 
+A node install normally includes npm as well, which might be a downgrade if you have upgraded npm separately. You can preserve your current npm and exclude it from the install:
+
+    $ n --preserve-npm 6.1.0
+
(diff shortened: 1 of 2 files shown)


JohnRGee:
Could also support `n --preserve-npm` for using with interactive version selection?

JohnRGee:
I was looking for issues around specific versions of Linux and found #100 due to `/usr/local/share/man` being a link on archlinux. Using tar for `--preserve-npm` extraction overwrites the link with a directory which is undesirable. I will have another go with rsync and see how that goes.

Issue #517, opened 2018-07-07
n latest gives Error:  invalid version


dhegglin:
# Issue Template:

### Describe issue including what OS you are using

docker 
root@642b8f51967f:/# uname -a
Linux 642b8f51967f 3.10.0-514.6.1.el7.x86_64 #1 SMP Wed Jan 18 13:06:36 UTC 2017 x86_64 GNU/Linux

### Describe what version of N you have

root@642b8f51967f:/# n --version
2.1.12

### Describe how you installed N

 RUN npm install --verbose -g n

### Steps to reproduce issue _1.2.3_

root@642b8f51967f:/# n

  Error: no installed version

### Describe the results you received

root@642b8f51967f:/# n 7.1.2

     install : node-v7.1.2

  Error: invalid version 7.1.2

### Describe the results you expected

install and use version 7.1.2

### What version of npm you are using

root@642b8f51967f:/# npm --version
6.1.0

### Additional information you deem important (e.g. issue happens only occasionally):

root@642b8f51967f:/# which n
/usr/bin/n
root@642b8f51967f:/# node -v
v10.6.0
root@642b8f51967f:/# curl --version
curl 7.52.1 (x86_64-pc-linux-gnu) libcurl/7.52.1 OpenSSL/1.0.2l zlib/1.2.8 libidn2/0.16 libpsl/0.17.0 (+libidn2/0.16) libssh2/1.7.0 nghttp2/1.18.1 librtmp/2.3
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IDN IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz TLS-SRP HTTP2 UnixSockets HTTPS-proxy PSL 
root@642b8f51967f:/# which curl
/usr/bin/curl
root@642b8f51967f:/# ls -l /usr/local/n/versions
total 0
drwxr-sr-x. 2 root staff 6 Jul  7 15:18 io
drwxr-sr-x. 2 root staff 6 Jul  7 15:18 node
root@642b8f51967f:/# ls -l /usr/local/n/versions/node
total 0
root@642b8f51967f:/# ls -l /usr/local/n/versions/io
total 0

dhegglin:
root@242d9dedbea0:/# n lts

  Error: invalid version 

root@242d9dedbea0:/# n latest

[...]

Issue #503, opened 2018-05-11
Feature Request: Add proxy support (for many of us behind the firewall)


stsvilik:
Would it be possible to add an option to specify a proxy when running `n`. You already doing it for `$HTTP_USER` and `$HTTP_PASSWORD`, so just need to extend `curl` params to accept `HTTP(s)_PROXY` as well.

Please :)

JohnRGee:
You may be able to use an environment variable recognised by `curl`, without changes in `n`? 

Either `https_proxy` or `HTTPS_PROXY`, depending on your implementation.

e.g. [How to use curl command with proxy username/password on Linux/ Unix](https://www.cyberciti.biz/faq/linux-unix-curl-command-with-proxy-username-password-http-options/)

stsvilik:
@JohnRGee I have those set in my environment - still no luck. Additionally my company uses middle-man certificate which requires me to ignore security when I use `curl`. Maybe a way to go about it is to have an optional configuration file for `n` (ex. `.nrc`) where such nuances could be configured?

JohnRGee:
I was wondering about certificates too as that seemed likely to be next problem! 
#324 #426

JohnRGee:
You may be able to use http for the traffic to the proxy, and avoid dealing with the middle-man certificate. In the examples my proxy server is `localhost:3128`. The default for curl is to use http to the proxy server and I am explicitly specifying that to make it clear that not using https for that part of the connection.

Using `curl` so you can see error messages if it fails, success looks like:

```bash
$ https_proxy='http://username:password@localhost:3128' curl --head https://nodejs.org/dist/
HTTP/1.1 200 Connection established
...
```

If that works then hopefully `n` does too:
```bash
https_proxy='http://username:password@localhost:3128' n ls
```

Tip: if you are using `sudo` for the install then it does not get your environment variables by default, and `sudo -E` preserves (passes) the environment variables.

stsvilik:
Unfortunately, neither `sudo -E n ls` nor explicitly setting proxy env in-line worked.
`curl` command only works like this `curl -k -x http://some.proxy.com:80 --head https://nodejs.org/dist/`

JohnRGee:
Thanks for info and example `curl` command. I understand possibilities better now.

A last question. Does a connection to the nodejs mirror on `http` work? e.g.

```bash
curl -x http://some.proxy.com:80 --head http://nodejs.org/dist/
[...]

Issue #430, opened 2017-04-14
How to set company proxy option when using n ?


daiando:
# Issue Template:
### Describe issue including what OS you are using
I'm using Centos 6.3 VM under company proxy.
I want to set proxy option when using 'n'.
I'm using npm with proxy settings but I can't find out how to use n with proxy option.
Do you have any idea ?

### Describe what version of N you have
2.1.7

### Describe how you installed N
npm install -g n

### Steps to reproduce issue _1.2.3_
1. n latest
2. Nothing has happen. (probably due to no proxy option)

### Describe the results you received
- prompt 

### Describe the results you expected
- upgrade node to latest version

### What version of npm you are using
4.5.0

### Additional information you deem important (e.g. issue happens only occasionally):

* Output of `which n:`
/usr/local/bin/n
* Output of `node -v:`
v.7.6.0

nineohnine:
+1 for this.  Looking for correct usage of `HTTP_PROXY`, `HTTPS_PROXY` env variables.

alqu:
I tried it successfully with setting the env-vars `http_proxy` and `https_proxy`.
If the proxy does not allow ssl tunneling then my PR #426 comes in handy.

daiando:
Oh, you're right.  I was able to use 'n' just like 'wget' or 'curl'.  Thank you for tips :)
`export HTTP_PROXY=http://your-proxy-url:port`
`export HTTPS_PROXY=http://your-proxy-url:port`
and then
`n latest` or whatever.

mmatalia:
sudo -E n latest
worked for me. -E option for sudo passes environment variables for current user

KuiWei004:
> Oh, you're right. I was able to use 'n' just like 'wget' or 'curl'. Thank you for tips :) `export HTTP_PROXY=http://your-proxy-url:port` `export HTTPS_PROXY=http://your-proxy-url:port` and then `n latest` or whatever.

worked for me !!!

KuiWei004:
> sudo -E n latest worked for me. -E option for sudo passes environment variables for current user

worked for me !!!

Issue #284, opened 2015-06-17
n fails to install anything


Ognian:
OS X 10.9
installed latest n 1.3.0 -> ```sudo npm install -g n```

```
sudo n node latest
cp: /usr/local/n/versions/node//bin: No such file or directory
cp: /usr/local/n/versions/node//lib: No such file or directory
cp: /usr/local/n/versions/node//share: No such file or directory
````
````
sudo n io latest
cp: /usr/local/n/versions/io//bin: No such file or directory
cp: /usr/local/n/versions/io//lib: No such file or directory
cp: /usr/local/n/versions/io//share: No such file or directory
```

notice the ```//``` before bin, lib, share
was using n 1.2.1 before without any problems
please help

Ognian:
see #250 
but uninstall install doesn't work for me...
``` n ls ``` doesn't show up anything ...
maybe this is a hint?

**Comment by maxrimue** [comment]
Does `n io ls` show anything?

**Comment by Ognian** [comment]
no

**Comment by Ognian** [comment]
OK, did some debugging:

the curl command issued by n and the reason for the empty list is: 

```
curl -# -L https://iojs.org/dist/
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.

```

maxrimue:
That's weird, curl has no problem with iojs.org's certificate on my machine, the npm version of n also works just fine

Ognian:
indeed, in safari the certificate of node and iojs are shown as trusted but curl don't think so...

Ognian:
OK, found it:
path contains another curl, not the OS X default one ...
let's see if everything works now...

haozxuan:
I have this problem too,can resolve?

Ognian:
Yes. everything works now...
So the only problem I see is that n doesn't show me the error from curl which would help in finding out the correct cause.

@haozxuan:
look with ```which curl``` at what curl is used and change your path to the correct one...

[...]

PR #524, opened 2018-08-12
Add --insecure (and proxy docs)

JohnRGee:
# Pull Request Template:

### Describe what you did

Added option for `--insecure` for working from behind a proxy which uses untrusted certificates. 

### How you did it

Separate commits so can cherrypick if desired:
1) Added `--insecure` option to n, and use it to disable certificate checks for curl/wget.
2) Added separate documentation on setting up proxy.

NB: wget options are still insecure by default. However, this change hopefully removes an obstacle to changing the wget default options in the future.

Code inspection that this is a new code path and will not affect previous use cases.

This option is partly to make it possible to use curl now when insecure is needed, and partly to allow making wget secure by default in future and bring it into line with curl.

Makes curl insecure opt-in, and not by default as suggested here: #323 #324 #426 

Make wget secure by default in future: #475 #509 

Other proxy related issues where `--insecure` could help: #284 #430 #503 #517

Running a local proxy with MITM certificates to demonstrate:

    $ export https_proxy=localhost:8080 
    $ n --lts
    $ n lts
    
    Error: invalid version 
    
    $ n --insecure --lts
    8.11.3
    $ n --insecure lts
    
         install : node-v8.11.3
           mkdir : /usr/local/n/versions/node/8.11.3
           fetch : https://nodejs.org/dist/v8.11.3/node-v8.11.3-darwin-x64.tar.gz
       installed : v8.11.3

Added: --insecure option to disable certificate checks

[Suggested addition]
Note: if curl is not present and `n` is using wget, the certificate check has historically been disabled and that behaviour has not changed yet. In the future this may change (and you can add the` --insecure` option now so will not be affected when it does change).

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -190,6 +190,7 @@ display_help() {
     -q, --quiet     Disable curl output (if available)
     -d, --download  Download only
     -a, --arch      Override system architecture
+    --insecure      Turn off certificate checking for https requests (may be needed from behind a proxy server)
 
   Aliases:
 
@@ -601,6 +602,18 @@ set_quiet() {
 }
 
 #
+# Set curl/wget to insecure mode
+#
+
+set_insecure() {
+  if command -v curl > /dev/null; then
+    GET="$GET -k"
+  else
+    GET="$GET --no-check-certificate"
+  fi
+}
+
+#
 # Remove <version ...>
 #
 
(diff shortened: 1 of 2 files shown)


JohnRGee:
Closing as:
* No feedback in 5 months, and would not have helped with issues reported in that period.
* A permanent solution for user is to create `.curlrc` or `.wgetrc` file.
* Potentially more useful to users would be to rework error handling so people can tell proxy and certificates are the problem in the first place!

PR #528, opened 2018-09-29
Hide cursor while selecting version from menu

shadowspawn:
# Pull Request Template:

### Describe what you did

Hide cursor while selecting node version from menu (i.e. `nvh` with no parameters).

### How you did it

Added calls to `tput` to hide and show cursor in `enter_fullscreen` and `leave_fullscreen`. These two routines already use `tput` to save and restore the screen contents.

Added redirection to suppress stderr just in case tput calls not supported by termcap. 

Deleted unused code to hide and show cursor using direct escape sequences.

Cosmetic only, not functional.
Tried the termcap calls on Mac High Sierra, Mac Mojave, Ubuntu, Centos, and archlinux.

On macOS there was a cursor left on screen below version list during version selection from menu, and sometimes a padlock would appear while moving selected version using arrow keys.

(graphic difference)

Hide cursor while selecting node version from currently installed versions using menu.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -121,12 +121,18 @@ set_arch() {
 #
 
 enter_fullscreen() {
-  tput smcup
+  # Set cursor to be invisible
+  tput civis 2> /dev/null
+  # Save screen contents
+  tput smcup 2> /dev/null
   stty -echo
 }
 
 leave_fullscreen() {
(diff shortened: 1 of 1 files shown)


shadowspawn:
Merged to develop for next release.

PR #532, opened 2018-10-16
Change which to command -v

JohnRGee:
# Pull Request Template:

### Describe what you did

Fix some issues on an OS without `which`, such as Centos docker image.

### How you did it

Changed `which` to `command -v` in the `check_current_version` routine.

Also added speech marks around variable dereference to fix error message that occurred before fixing above.

Checked `n` and `n prune` on MacOS and Centos.

Before (with no `which` command):
```bash
# n prune
/usr/local/bin/n: line 263: which: command not found
/usr/local/bin/n: line 263: which: command not found
/usr/local/bin/n: line 625: [: node/10.12.0: unary operator expected
```

After:
```bash
# n prune
```

Remove use of `which` and use generally available `command -v`.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -262,7 +262,7 @@ check_current_version() {
     for bin in ${BINS[@]}; do
       if diff &> /dev/null \
         $BASE_VERSIONS_DIR/$bin/$current/bin/node \
-        $(which node) ; then
+        $(command -v node) ; then
         active=$bin/$current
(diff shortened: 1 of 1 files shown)


troy0820:
This looks good.  I'm going to try to pull it in today but trying to do them one at a time is interesting with the order I need them to work.

PR #535, opened 2018-10-30
Improved timeliness of LTS changes across major versions

pdspicer:
### Describe what you did
Changed display_latest_lts_version to use index.tab file for retrieving the lts version

### How you did it
Added /index.tab to path for GET request to mirror, piped output to locate only "lts" rows, took only the first one, discarding header row.

Modified output to conform with previous style (specifically, removing the "v" prefix from version number, output matches existing functionality so should not break existing dependents).

N/A

N/A

Completed prior to opening PR

Changed display_latest_lts_version to use index.tab file for retrieving the lts version

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -723,16 +723,12 @@ display_latest_stable_version() {
 
 display_latest_lts_version() {
   check_io_supported "--lts"
-  local folder_name=$($GET 2> /dev/null ${MIRROR[$DEFAULT]} \
-    | egrep "</a>" \
-    | egrep -o 'latest-[a-z]{2,}' \
-    | sort \
-    | tail -n1)
-
-  $GET 2> /dev/null ${MIRROR[$DEFAULT]}/$folder_name/ \
-    | egrep "</a>" \
-    | egrep -o '[0-9]+\.[0-9]+\.[0-9]+' \
-    | head -n1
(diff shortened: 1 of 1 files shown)


JohnRGee:
(I am not a maintainer.)

A couple of questions in case you know:

1) How much difference does this make to the timeliness? (i.e. how much sooner does a version show up in index.tab than in the html files?)

2) Does index.tab get updated just once for a release, or multiple times as the various architectures get built? (I am concerned whether some architectures won't have a download available when first added to index.tab. The architectures are included in index.tab so support for checking this could be added if needed.)


And a couple of comments.

I actually like the idea of using index.tab so using the file intended for automated consumption rather than the file intended for humans. I think if make this change for LTS then the version number lookup should also change, rather than have two approaches.

Some [n history](https://github.com/tj/n/pull/322#issuecomment-156885208) from when lts support was added: apparently index.tab was used in a (very) early version, but wasn't always up to date.

shadowspawn:
I am intending to move `n` to using `index.tab`, but as full replacement for version scanning.

pdspicer:
Hey, apologies for not getting back to this, totally agree with everything you were saying above and happy to lend a hand if desired (don’t know if anything from this PR might actually be of use to you..). Regarding timeliness, I think we’re talking about a difference of hours, I am just impatient :D

Issue #441, opened 2017-06-08
Uninstalling n on OSX


mdanetzky:
I'm using MacOS Sierra and n 2.1.4 installed using homebrew. To have a clean uninstall I need to remove all installed versions of node. I can't do that, because n rm tells me that
```
Error: cannot remove currently active version
```
It would be great do have an option to deactivate n (just like in nvm).

troy0820:
Could you use the template so I can understand where everything is?  I honestly wouldn't use brew to install n, as it's not in the documentation.  Brew installs in /usr/local/Cellar and symlinks it.

mdanetzky:
I used:
```
brew install n
```
Unfortunately every npm execution needs elevated privileges - which is the reason why I want to remove it. Nonetheless I can imagine, that a functionality to deactivate n might be useful in other scenarios too. For example while switching between various installations.

PS. I just used sudo rm to remove everything that resembled node, npm or node_modules from my system.

troy0820:
Uninstall with brew and try to install n with the preferred method in the readme.  If you don't own the directory where n installs, you will either have to chown them or run sudo everytime you run it.

Issue #400, opened 2016-12-18
n selector doesn't work


af-fess:
I installed `n` with `npm install -g n`

when I run `$ n` I get output:

```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]

    node/4.4.7
    node/6.6.0
    node/7.2.1
```

when I run:

`$ n 4.4.7` - nothing happens. No output, `node` version doesn't change.

tried to reinstall `n` - same thing. Is there any  `--verbose` flag?   

```
 $ node -v
v7.1.0
```
Any ideas?

dcsan:
I get exactly this too. 
using zsh

➜  ~ n --version
2.1.3

the up/down arrows on the selector don't work, and there is no arrow to indicate currently selected version.


I seem to have other issues with the node installation.
what is the best way to completely uninstall n?

digitalmaster:
Same issue here

matter3:
@digitalmaster thanks, fixed for me as well.

gogoout:
+1
using zsh.
I have no `nvm` installed
```
 ~  nvm
zsh: command not found: nvm
```

dcsan:
could be related to zsh, i also have the issue and am using zsh.

andresmihkelson:
I am using bash and with the latest n, the same happens. No nvm also. Using Fedora 24. Also, after running n fails, I cannot type anything into the prompt any more.

maciekrb:
Make sure you uninstall your previous version of node. For example, if you are using OSX and MacPorts or `brew`, make sure `which node` points to `/usr/local/bin/node` (or the path you chose to install `n` node).

troy0820:
The problem is that it can't find the `$versions_paths` directory.  This could be because `find` isn't installed or the `$N_PREFIX` isn't pointing to where the versions are installed.  

@andresmihkelson Are you using a container or is this local?  Make sure you have the `findutils` package.  I have a container that you can test this on, if you have docker.

```bash
docker run troy0820/node-n-fedora
npm i -g n
n 4.6.0
n lts 
n latest
```
Install a couple of versions and then switch with `n` and you should see it work fine.

andresmihkelson:
[...]

Issue #391, opened 2016-10-28
Installing a -test version bugs out grep.


EvanCarroll:
Grep is reading the search string when you [install nightly](https://github.com/tj/n/issues/376) as an argument. Using the workaround in #376 https://github.com/tj/n/issues/376#issuecomment-252156076,

```
$ node --version
v7.0.0-test2016100609987d242b

```

gives you grep errors.

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

Issue #367, opened 2016-07-29
Can't select different node versions with node 6.x on Ubuntu


valgaze:
@qw3rtman 
To add another data point to the pile, using a Bitnami AMI (basically a batteries-included Linux image for quick deployments) on Amazon Web Services with node 6.2.1 as default exhibits behavior similar to the kind described in **Issue #355** The solution proposed by @ElegantSudo did not resolve the issue. 

**Bare minimum number of steps to reproduce here: https://gist.github.com/valgaze/728a9758d6abd73702ef27f7986f4b59**
- Image: http://aws.amazon.com/marketplace/pp/B00NO1HJ56?ref=cns_srchrow [Linux/Unix, Ubuntu 14.04.3 | 64-bit Amazon Machine Image (AMI) | Updated: 7/7/16]
- OS:  Ubuntu 14.04.3
- Running `$ uname -or`:  3.13.0-87-generic GNU/Linux
- Starting node version (preinstalled on image): 6.2.1

Here's a gif the wacky behavior described:

![1_weirdn](https://cloud.githubusercontent.com/assets/1396559/17233170/6976557e-54e2-11e6-92c6-851ae6ed4af5.gif)

Will post a gist of replication instructions in a moment: https://gist.github.com/valgaze/728a9758d6abd73702ef27f7986f4b59

valgaze:
@qw3rtman As of this morning, SSH'd in & it works like a charm

valgaze:
Very strange-- just tried again this AM and it seems to work perfectly

valgaze:
This only seems to occur on 6.2.1, for some reason seems OK on 6.3.1

hk0i:
Seeing similar behavior on node `v7.10.0`,

Pressing the <kbd>up</kbd> key reveals the grep usage message as shown in the image above:

```
Usage: grep [OPTION]... PATTERN [FILE]...
Try 'grep --help' for more information.

    node/0.10.31
    node/8.2.1
```

kmsheng:
This also occur on node `v8.4.0`

garuda:
Nothing on this huh? I'm seeing the same :( Uninstalled, reinstalled, nadda

adrianlyjak:
same thing on macos 10.11.6, n version 2.1.7. Using up and down arrows to select version just outputs `grep` help documentation:

```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
[...]

PR #541, opened 2018-12-04
Fix menu selection when installed version not present

shadowspawn:
# Pull Request Template:

### Describe what you did

Fixed menu selection when active node version is not present in list. (#292 #367 #400)

Removing prior work-around in PR #296 which blocked calling `n rm` on the active version, but that causes some confusion when people trying to uninstall. (#169 #327 #441) 

### How you did it

Added speech marks around variable dereference in prev/next called when moving menu selection up or down.

Deleted now unnecessary lines of code blocking removal of current version.

Code inspection (simple changes). 
Verify menu selection still works using `n`.
Remove a cached version using `n rm <version>`.

Solves #292 #367 #391 #400

```bash
$ n 4.9.1
$ n 8.14.0
$ rm -rf /usr/local/n/versions/node/8.14.0
$ n 
(hit up or down arrow)

usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]

    node/4.9.1
```

Menu selection works in this situation after fix. Hitting the up/down arrow from first display moves selected item to top/bottom version, with no error message displayed.

Fix menu selection of cached node versions when active node version is not present in cache. Allow removal of active node version from cache.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -227,7 +227,7 @@ show_cursor() {
 #
 
 next_version_installed() {
-  list_versions_installed | grep $selected -A 1 | tail -n 1
+  list_versions_installed | grep "$selected" -A 1 | tail -n 1
 }
 
 #
@@ -235,7 +235,7 @@ next_version_installed() {
 #
 
 prev_version_installed() {
-  list_versions_installed | grep $selected -B 1 | head -n 1
+  list_versions_installed | grep "$selected" -B 1 | head -n 1
 }
 
 #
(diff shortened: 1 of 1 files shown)


JohnRGee:
The active node version missing from the menu list can also happen when the active version of node was installed without using `n`.

JohnRGee:
(PR https://github.com/tj/n/pull/489 was fixing same issue but with a focus on the cause of the situation.)

JohnRGee:
(PR #390 was fixing same issue, with trying nightly as the cause.)

PR #551, opened 2019-01-20
installation folder became dynamic

m7md2112:
# Pull Request Template:

### Describe what you did
 make installation directory auto detected.

### How you did it

grep current **npm**  directory  using `which npm`

firstly I manually installed node.js in this directory `/home/myUserName/.local` because I don't want run **sudo** every time when I need to install a global package so [n](https://github.com/tj/n) didn't work with me. by review the code I found the installation directory assigned statically to "N_PREFIX" variable, so I assigned this command result to it and _voila_  finally worked.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -5,7 +5,7 @@
 #
 
 VERSION="2.1.12"
-N_PREFIX=${N_PREFIX-/usr/local}
+N_PREFIX=`which npm  | rev | cut -c9- | rev`
 BASE_VERSIONS_DIR=$N_PREFIX/n/versions
 
 #


JohnRGee:
1) Your patch assumes that `which npm` is known. However, `n` can be used without either `npm` or `node` installed. The install location defaults to `/usr/local` and is customised by defining `N_PREFIX` (as you found).

2)  Installing to a custom location is fully supported. You need to define `N_PREFIX` yourself. You put two lines in your shell initialisation file. The first to specify the prefix for where to install, and the second to add the location to `PATH` so your shell can find `npm` and `node`. You probably have a line like the second already. e.g.

    N_PREFIX=/home/myUserName/.local
    PATH=${N_PREFIX}/bin:$PATH

The `node` binary will be installed to:

    /home/myUserName/.local/bin/node

Issue #545, opened 2018-12-22
Code Suggestion: add testing framework and tests


shadowspawn:
# Problem

`n` does not have any tests for contributors to run after making changes, to add to when making changes, or to prevent regressions. I would like the reassurance that the tests still pass when making a Pull Request, or similarly when reviewing a Pull Request. (End users would hopefully benefit from the increased checks with more robust releases, but would not run the tests themselves.)

The Pull Request template asks a good question and I don't think there is currently a good answer:

> How to verify it doesn't effect the functionality of n

# Possible Solutions

I looked around and found [bats-core](https://github.com/bats-core/bats-core), which is quite widely used. I liked the strong focus on bash, and test and output formats

Using a caching proxy is an optional optimisation for the multiple web requests. On Mac, I have used [mitmproxy](https://mitmproxy.org)

Using `docker` allows running commands or test suite on multiple flavours of operating system, all self-hosted.

# Feedback

I am interested in any comments (including of course from maintainers!). 

I develop primarily on a Mac, and use Linux flavours in VMs and containers as needed but not as my desktop, so found technology that works well for my situation. I know @troy0820 uses docker, and might be on a Mac too. There may be other frameworks or approaches that would work better for `n` contributors with other environments?

A framework for providing and running tests is separate from writing the tests themselves. Any comments on good tests also welcome.

I am aware I don't need to get this into `n` to use, as I can make a stand-alone repo for hosting the tests. I have used `bats-core` and `mitmproxy` and `docker` together to write [tests for nvh](
https://github.com/JohnRGee/nvh/blob/master/docs/tests.md) so know I can get something functional. 

# History

Having some tests gets mentioned as desirable in #246. A roll-your-own framework and a single test was included with PR #364.

troy0820:
Definitely, agree.  I want to automate a lot of the testing because what ends up happening is that I make a bunch of containers with Docker to test the functionality of it.  If we can get something started, I'm happy with the suggestions here.

JohnRGee:
Started work in: https://github.com/JohnRGee/n-test

JohnRGee:
An update. I spent some time on this at start of year, and have a functional setup and initial suite of tests in https://github.com/JohnRGee/n-test. Can open a PR when wanted.

 I have ideas for improvements and additions but a low priority at the moment and I am not planning to change approach significantly (despite #554 #545). Easy to add travis support if desired.

Infrastructure:
- tests can be run on native OS, independent of developers own installs and configuration
[...]

PR #555, opened 2019-02-07
I Added some basic tests via travis yml to make sure it still works as expected

frank-dspeed:
# Pull Request Template:

### Describe what you did
Added Basic Dockerfile for tests and also Production use inside docker
Added .travis.yml to enable this repo inside travis-ci for automated tests on push

### How you did it
i writed into the Dockerfile the needed lines and added also some tests inside .travis.yml

you can verify that it doesn't affect the functionality of n via n did not change :+1: 

solves issue #545 #555

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - docker
 
 before_install:
-  - docker build -t dockerimages/n .
+  - docker build -t n .
 
 script:
 #  - docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "apt update && apt install -y curl && n latest && node -v"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: trusty
+
+services:
+  - docker
+
+before_install:
+  - docker build -t dockerimages/n .
+
+script:
+#  - docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "apt update && apt install -y curl && n latest && node -v"
+- docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "n 0.8 && node -v"
+- docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "n latest && node -v"
+- docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "n stable && node -v"
+- docker run -it --rm -v $PWD/bin/n:/usr/bin/n n /bin/sh -c "n 8 && node -v"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM ubuntu:18.04
+RUN apt update && apt install -y curl
+ADD ./bin/n /usr/bin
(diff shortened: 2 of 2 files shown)


frank-dspeed:
Closes: #555 #545

JohnRGee:
1) The docker build generates warnings about using `apt` in scripts:

`WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`

2) I don't think you need both volume mount at runtime and `ADD` at build time. Either will make `n` available in the container.

3) The tests are pretty simple, just testing that node runs after an install, not even that it is the expected version. I don't think it comes close to covering what #545 proposes..

frank-dspeed:
@JohnRGee ok you want to have it test for version no problem you get it :) you want to complain about apt and not apt-get ok you get apt-get :D

JohnRGee:
I offered some feedback on your PR, but you should decide what are worthwhile improvements.

You stated your PR "solves" the issue I opened (#545), and added that this "closes" my issue. I personally disagree. (I still welcome feedback on my own proposal and work, outside this PR.)

frank-dspeed:
@JohnRGee we can simply add every tests that you like i don't have a plan for tests but i looked at your examples and i simply don't understand them :)

i will get this one merged and if you like you can base your tests on this simply remove the bats dependency as this is not needed we can test bash with bash. but i think we need no additional test runner simply a collection of docker run commands should solve all test needs. we can design this commands so that they error.

i also see no need for this proxy stuff and that i never tested something with a proxy in between that adds extra point of error

frank-dspeed:
@JohnRGee you could really help me with some valid test cases in a formart like
- Should return right version when "n latest"
- Should install in directory x when "PREFIX=x n latest"

JohnRGee:
proxy: I made the proxy entirely optional. Why even bother? Big performance boost for running tests locally, with full integration tests that do downloads. I did a check today and my tests ran in 1 min with proxy, 28 min without. node.js was so slow today!

bats: not essential. I wanted to try a test runner and structured tests and focus on the other challenges. Planning ahead for lots of tests.

I like the elegance of your idea of "simply a collection of docker run commands" in theory, but yet to see if it can cope with a variety of tests. A case you may not have considered is running tests on macOS which is not available as a container. (macOS is available on Travis as a host OS.)

I think some important test cases are:
- `n lts` installs expected version of node
- `n latest` installs expected version of node
- `n x.y.z` (explicit version) installs expected version of node
- `n --lts` outputs expected version of node
- `n --latest` outputs expected version of node
- installing with defined N_PREFIX installs to N_PREFIX

At least some (or all) tests should be run with curl and with wget. (Containers to the rescue!)
At least some (or all) tests should be run on MacOS and on Linux.

frank-dspeed:
@JohnRGee i agree on running tests localy but i think still inside docker no one wants to mess up his system only alternativ that i see is using a vm for local testing.

In General i think its ok if we activate travis and then if some one pushes code the tests are running for every pull request and that should be enought for him.

shadowspawn:
@frank-dspeed Do you want to take this any further?

shadowspawn:
Going with #545 approach.

Issue #570, opened 2019-06-19
Version selection logic fails with small/old version numbers


DeeDeeG:
### Describe issue including what OS you are using

Can't install old versions (0.7 or older)

(Using Ubuntu 19.10...)

### Describe what version of N you have

4.1.0, but also happens on develop branch (5.0.0-next.0)

### Describe how you installed N

I cloned https://github.com/tj/n, and put a symlink to the n binary in my /usr/local/bin folder:
`sudo ln -s /home/[me]/n/bin/n /usr/local/bin/n`.

This way I can switch n versions by checking out different git branches in my local n repo.

I also made sure n versions are installed within my home directory:
`export N_PREFIX=/home/[me]/n-prefix`

### Steps to reproduce issue

1. `n 0.0.1` _(also applies when running `n 0.0.2`, all the way through `n 0.0.6`)_
2. Error: invalid version '0.0.1'

---

1. `n 0.1`
2. _node version 0.12.18 is installed instead_

---

1. `n 0.1.0` _(also applies when running `n 0.1.1`, all the way through `n 0.1.104`)_
2. Error: invalid version '0.1.1'

---

1. `n 0.2` _(also applies when running `n 0.3`, all the way through `n 0.7`)_
2. Error: invalid version '0.2'

---

1. `n 0.8.1`
2. _node version 0.8.19 is installed instead_

--

1. `n 0.8.2`
2. _node version 0.8.28 is installed instead_

--

1. `n 0.8.3` _(also applies when running `n 0.8.4`, or `n 0.8.5`)_
2. Error: invalid version '0.8.3'

--

1. `n 1`
2. _node version 12.4.0 is installed instead_

### Describe the results you received

Some older versions are considered invalid... even though it doesn't say these old versions aren't supported by n.

And in some cases, small numbers are interpreted/expanded unexpectedly: "1" becomes "12" or "19", 2 becomes "28".

### Describe the results you expected

Old versions should download and install without problems, OR they should be expressly not supported (ideally with documentation to go along with it, maybe just a mention in the README).

Weird corner case with `n 0.8.3` through `n 0.8.5` should be handled.

Also, small numbers like "1" or "2" shouldn't be interpreted/expanded as "12" "19" or "28".

### What version of npm you are using

I'm not invoking/using npm here

### Additional information you deem important (e.g. issue happens only occasionally):

My active version of node always comes up as 10.16.0, which I have already installed via nvm... I don't think that changes any of the issues described, however.

[...]

PR #572, opened 2019-06-21
bugfix: Improve install logic (regex fixes)

DeeDeeG:
# Pull Request Template:

### Describe what you did
Fix the `install()` function to install the expected version when doing `n [version]` (see #570 for details)

### How you did it

[Updated 28 June] Adds a trailing dot toward the end of the grep filtering steps in `install()`.

`n [version]` for all the versions mentioned in #570 

solves issue #570 

#### Before:

```
$ n 0.8.1

  installing : node-v0.8.19
       mkdir : /home/mate/n-prefix/n/versions/node/0.8.19
       fetch : https://nodejs.org/dist/v0.8.19/node-v0.8.19-linux-x64.tar.gz
   installed : v0.8.19 to /home/mate/n-prefix/bin/node
      active : v10.16.0 at /home/mate/.nvm/versions/node/v10.16.0/bin/node
```

#### After:

```
$ n 0.8.1

  installing : node-v0.8.1

  Error: invalid version '0.8.1'
```

(see "Describe what you did" above)

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -533,7 +533,7 @@ install() {
       | grep -E -v '^0\.[0-7]\.' \
       | grep -E -v '^0\.8\.[0-5]$' \
       | sort -u -k 1,1n -k 2,2n -k 3,3n -t . \
-      | grep -E "^$version" \
+      | grep -E "^$version\." \
       | tail -n1)"
 
     test "$version" || abort "invalid version '${1#v}'"
--- a/bin/n
+++ b/bin/n
@@ -525,7 +525,7 @@ install_lts() {
 install() {
   local version=${1#v}
 
-  local dots="${version//[^.]*/}"
+  local dots="${version//[^.]/}"
   if test ${#dots} -lt 2; then
     version="$($GET 2> /dev/null "${MIRROR[DEFAULT]}" \
       | grep -E "</a>" \


shadowspawn:
I am interested in fixing the "dots" test and making the partial number lookup match whole numbers, but I don't want to change the program flow at the same time. 

If you want to simplify this down to just those changes (which will be tiny!), I'll take another look.

DeeDeeG:
Wow, this really did simplify down. It's all of three characters of diff!

I went through a lot of iterations of working this out in my mind, tweaking the script, trying things, and in the end it ended up being pretty similar to the starting point.

I was even going to change the `case` to an if-then. I would plan to do it that way if we decide to filter "more than two dots" versions through the grepping process. (Which, if anyone else is reading this after the fact, was how I initially submitted this PR.)

DeeDeeG:
I can squash these together if you prefer.

shadowspawn:
Thanks for changes, and for squash offer. I can do the squash when I merge so you don't need to thanks.

shadowspawn:
FYI: I am planning to include this in v5. I am leaving the `@next` version up for a few weeks and working on another project. It might be a week or two before I test and merge this but I don't anticipate any problems.

shadowspawn:
Shipped in v5.0.0

PR #575, opened 2019-07-30
No check certificate for curl

Bacra:
No check certificate for `curl` as same as `wget`

https://unix.stackexchange.com/questions/60750/does-curl-have-a-no-check-certificate-option-like-wget

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -49,7 +49,8 @@ fi
 # Ensure we have curl or wget support.
 #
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
Actually, it is the `wget` usage in `n` that I intend to change, and have it check certificates by default. 

See: #324 #475 #509 #524

If you need to disable certificate checking, then using a `.curlrc` file is an option.

Issue #377, opened 2016-09-30
Allow switching between local versions using major version number when internet is not available


Corluhan:
Say I have 2 versions installed:

```
    node/0.12.16
  ο node/6.7.0
```

I can easily switch between them when specifying the full version number:

```
$ sudo n 0.12.16
$ sudo n 6.7.0
```

This doesn't work with the major numbers:

```
$ sudo n 6.

  Error: invalid version 6.

```

The ideal behavior is to select the latest installed version with the specified major version number.

qw3rtman:
@Corluhan: Is this only an issue when you have no Internet connection?

shadowspawn:
This is an interesting idea, but would need a different approach to clarify the two different behaviours of picking latest local version or latest downloadable version.  

A simple workaround is to use the menu selection and pick from the installed versions.

This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

shadowspawn:
Added --offline option in n v9.2.0

PR #585, opened 2019-08-25
Adding n latest_installed / n which latest_installed to get latest locally installed versions

techiev2:
# Pull Request

Adding n latest_installed / n which latest_installed to get latest locally installed versions

## Problem

n latest resolves against remote and does not allow for local latest versions to
be used with a shortcut. [1](https://mobile.twitter.com/techieV2/status/1159866059534393344)

## Solution

There are existing functions that give the latest installed local versions but is not exposed via the command line. This change adds a couple of helper functions that use the existing list version function footprints to get the latest installed NodeJS version/binary path.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -261,10 +261,12 @@ Commands:
 
   n                              Display downloaded node versions and install selection
   n latest                       Install the latest node release (downloading if necessary)
+  n latest_installed             Get the latest installed node version
   n lts                          Install the latest LTS node release (downloading if necessary)
   n <version>                    Install node <version> (downloading if necessary)
   n run <version> [args ...]     Execute downloaded node <version> with [args ...]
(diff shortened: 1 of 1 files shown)


shadowspawn:
On a related note, there have been issues suggesting version selection with no internet: #377 #404

This PR is headed in that direction, but won't cope with possible situations. (Try installing `rc/latest` for example.) 

For syntax, the idea for a more general solution that comes to mind is an option flag, perhaps "local" or "cached" or "offline". `n ls` has a local context so makes sense to list a major version number, but might be bit ambiguous for latest. 

```
n ls latest
n --offline latest
n ls 8
n --offline 8
```

I suggest opening an issue if you want to discuss this further, or proposing a particular aspect (like adding new label for `latest_installed`) to see if there is interest from others.

Issue #139, opened 2013-07-09
add support for "x", "d", or delete to remove a selected version


tj:
in the `n` interactive listing

tedgaydos:
I am going to look into adding some more features to the interactive menu.

maxrimue:
@tedgaydos Would we want to add this to 1.x.x too? Or is the branch just for bug fixes?

shadowspawn:
This issue has not had any activity in over six months. It isn't likely to get acted on due to this report.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

PR #590, opened 2019-10-10
Press "d" to remove version in menu

thomasruiz:
# Pull Request

## Problem

It is quite hard to remove a bunch of previously installed versions without using prune. If I want to keep multiple versions, I cannot do that easily.

## Solution

You can now press `d` to remove versions in the menu. You cannot remove the currently active node.

## ChangeLog

You can now press `d` to remove versions in the menu. You cannot remove the currently active node.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -384,18 +384,13 @@ display_versions_paths() {
 #
 
 display_versions_with_selected() {
-  set_active_node
   selected="$1"
   echo
   for version in $(display_versions_paths); do
-    installed=""
-    if test "$version" = "$g_active_node"; then
-      installed=" [active]"
-    fi
     if test "$version" = "$selected"; then
-      printf "  ${SGR_CYAN}ο${SGR_RESET} %s${SGR_RED}%s${SGR_RESET}\n" "$version" "$installed"
+      printf "  ${SGR_CYAN}ο${SGR_RESET} %s\n" "$version"
     else
-      printf "    ${SGR_FAINT}%s${SGR_RESET}${SGR_RED}%s${SGR_RESET}\n" "$version" "$installed"
+      printf "    ${SGR_FAINT}%s${SGR_RESET}\n" "$version"
     fi
   done
   echo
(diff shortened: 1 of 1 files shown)


shadowspawn:
Saw suggested in this old issue: #139

shadowspawn:
Is there a reason to prevent deleting the active version? It complicates the code and description, and I don't think is needed. (In older versions of `n` there was some code to prevente deleting the active version using `n rm` because it caused drawing errors in the menu, but the drawing problems have been fixed.)

bin/n
```diff
@@ -384,17 +384,22 @@ display_versions_paths() {
...
-      printf "  ${SGR_CYAN}ο${SGR_RESET} %s\n" "$version"
+      printf "  ${SGR_CYAN}ο${SGR_RESET} %s${SGR_RED}%s${SGR_RESET}\n" "$version" "$installed"
     else
-      printf "    ${SGR_FAINT}%s${SGR_RESET}\n" "$version"
+      printf "    ${SGR_FAINT}%s${SGR_RESET}${SGR_RED}%s${SGR_RESET}\n" "$version" "$installed"
     fi
   done
   echo
-  printf "Use up/down arrow keys to select a version, return key to install, q to quit"
+  printf "Use up/down arrow keys to select a version, return key to install, d to remove (only inactive versions), q to quit"
```
shadowspawn: Suggest "d to delete" so the shortcut and action match.

shadowspawn:
(Disclaimer: I have not tried running the code yet! Opening the discussion after reading the code. Thanks.)

thomasruiz:
> Is there a reason to prevent deleting the active version? It complicates the code and description, and I don't think is needed. (In older versions of `n` there was some code to prevent deleting the active version using `n rm` because it caused drawing errors in the menu, but the drawing problems have been fixed.)

Well, there's a bit of a bug when you delete the active version, as you can see on the gif below. After I delete node 10 (which is the active version), no version is selected, and you have to press down or up two times to get to the top or bottom of the list. However, I can change the code to select the previous/next version automatically.

Besides, if you delete all the versions, you end up on an empty screen. Then again, the code can be changed to exit when there is no installed version anymore.
n-delete-current-version

shadowspawn:
Ah, I was wondering what the post-delete behaviour was, now I see the 

```
display_versions_with_selected "${g_active_node}"
```

> However, I can change the code to select the previous/next version automatically.

I _think_ that is going to be nicer all the time. If for example you are deleting a bunch of 8 versions now 10 is out. Scroll to the 8s and hit `d` a few times and don't have to keep scrolling after each delete to get back to the old versions.

thomasruiz:
@shadowspawn It's all done. The UX is much better now!

I can rebase this PR to contain only 1 commit if you want.

shadowspawn:
I gave it a bit of a try and liking it. I'll merge it in sometime this week. Thanks.

shadowspawn:
FYI @thomasruiz: I am planning to do a release this weekend with this new feature.

shadowspawn:
Released 6.1.0. Thanks @thomasruiz for your contributions.

PR #597, opened 2019-11-13
Use proper path to store application data

anddam:
# Pull Request

## Problem

By default n is installing data in `$N_PREFIX/n`, with default value this means `/usr/local/n` rather than a more appropriate `/usr/local/share/n`, cf. hier(7)

## Solution

Change `$CACHE_DIR` default.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -45,7 +45,7 @@ VERSION="6.0.1"
 N_PREFIX="${N_PREFIX-/usr/local}"
 N_PREFIX=${N_PREFIX%/}
 readonly N_PREFIX
-readonly CACHE_DIR=$N_PREFIX/sharen/versions
+readonly CACHE_DIR=$N_PREFIX/share/versions
 
 N_NODE_MIRROR=${N_NODE_MIRROR:-${NODE_MIRROR:-https://nodejs.org/dist}}
 N_NODE_MIRROR=${N_NODE_MIRROR%/}
--- a/bin/n
+++ b/bin/n
@@ -45,7 +45,7 @@ VERSION="6.0.1"
 N_PREFIX="${N_PREFIX-/usr/local}"
 N_PREFIX=${N_PREFIX%/}
 readonly N_PREFIX
-readonly CACHE_DIR=$N_PREFIX/n/versions
+readonly CACHE_DIR=$N_PREFIX/sharen/versions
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
1) You have a typo in the path you suggested in the Pull Request.

2) It is easy to change the location for new users, but hard to manage the transition for people upgrading from an earlier version with files in the old location, so changing the location is a major step.

3) I don't think `share` is appropriate. From http://man7.org/linux/man-pages/man7/hier.7.html

> `/usr/local/share` 
> Local application data that can be shared among different architectures of the same OS.

but the extracted tarballs stored under `n/versions` are architecture specific.

anddam:
> 1. You have a typo in the path you suggested in the Pull Request.

Wow, and it was a single word commit.

> 2. It is easy to change the location for new users, but hard to manage the transition for people upgrading from an earlier version with files in the old location, so changing the location is a major step.

If you want to keep compatibility I suggest to detect a setup with the previous path then either print out a message to the user and have him migrate the versions, or provide the migration for him.
The migration could be either just move the folder or installing using the same version.

> 3. I don't think `share` is appropriate. From http://man7.org/linux/man-pages/man7/hier.7.html
> > `/usr/local/share`
> > Local application data that can be shared among different architectures of the same OS.
> but the extracted tarballs stored under `n/versions` are architecture specific.

I am not sure how to interpret that sentence, that refers to system's application whereas n is acting like a package manager of its own.
I would be fine with `/usr/local/share` given what I see used most of time and [XDG specification for XDG_DATA_DIRS](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), or if the architecture thing bothers you, `/opt/n/`.

shadowspawn:
I did some more reading and found another detailed description, with an additional mismatch of the intended use of `/usr/local/share` and what `n` uses `/usr/local/n` for.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html

Emphasis added:

> Any program or package which contains or requires data that *doesn't need to be modified* should store that data in /usr/share (or /usr/local/share, if installed locally). It is recommended that a subdirectory be used in /usr/share for this purpose. Applications using a single file may use /usr/share/misc.

The directory n creates has downloads of node added and removed by various commands. So it is not a read-only folder, and not matching the described use of `/usr/share` and `/usr/local/share`.

I was a little surprised myself when I first saw that `n` was writing to `/usr/local/n`, but each time I have read about the system folders I have not found the perfect place and moving the folder is a difficult transition. Status quo rules for now.

PR #614, opened 2020-03-05
fix arch match rule for armv8l

yunyu950908:
# Pull Request

## Problem

I have tried to run `n` on my armv8l arch device, but could not get an appropriate result.

```
➜  ~ n lts
  Error: no version found for 'lts'
```

```
➜  ~ n lsr
Listing remote... Displaying 20 matches (use --all to see all).
➜  ~
```

After doing some research I saw some information about [the relations among armv7l, armv8l, and arm64 from StackOverflow](https://stackoverflow.com/questions/41091934/are-armv8-and-arm64-the-same):

> "arm64" represents the AArch64 state of the ARMv8-A architecture; there is no "armv8" target.

So, here is the problem in the script

https://github.com/tj/n/blob/a0cfc8079126c84125bbd53ad4f62e2406bb8e98/bin/n#L853-L861

## Solution

Just declare `armv8l` use `arm64` arch

```
aarch64 | armv8l) arch=arm64 ;;
```

And then it works
```
➜  n git:(master) ✗ ./bin/n lts

  installing : node-v12.16.1
       mkdir : /root/n/n/versions/node/12.16.1
       fetch : https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-arm64.tar.xz
####################################################################      95.7%
```

```
➜  n git:(master) ✗ ./bin/n lsr
Listing remote... Displaying 20 matches (use --all to see all).
13.10.1
13.10.0
13.9.0
13.8.0
13.7.0
13.6.0
13.5.0
13.4.0
13.3.0
13.2.0
13.1.0
13.0.1
13.0.0
12.16.1
12.16.0
12.15.0
12.14.1
12.14.0
12.13.1
12.13.0
```

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -853,7 +853,7 @@ function display_tarball_platform() {
   case "${uname_m}" in
     x86_64) arch=x64 ;;
     i386 | i686) arch="x86" ;;
-    aarch64) arch=arm64 ;;
+    aarch64 | armv8l) arch=arm64 ;;
     *)
       # e.g. armv6l, armv7l, arm64
       arch="${uname_m}"


shadowspawn:
I did some research but have not find a good explanation of what `arm8l` means. I did find two references that say arm8l does not guarantee 64 bit support, which makes it harder to test for arm64 support accurately for `arm8l`:

- https://github.com/bannsec/revenge/issues/3
- https://github.com/void-linux/xbps/issues/43

I see there are two armv8 variations which do not include 64 bit support, but I am not sure if they would be identified with `arm8l`:

- https://en.wikipedia.org/wiki/ARM_architecture#ARMv8-R_and_ARMv8-M

Without making changes to `n`, I'll note there is support to override the architecture detection with `--arch`.

shadowspawn:
What is the full `uname` string from your armv8l arch device?

```
uname -a
```

yunyu950908:
```
Linux localhost 3.18.66-perf+ #1 SMP PREEMPT Fri Sep 14 21:27:03 CST 2018 armv8l GNU/Linux
```

PS: I noticed that the download link of tarball for armv8 on https://nodejs.org/en/download/ is this https://nodejs.org/dist/v12.16.1/node-v12.16.1-linux-arm64.tar.xz

shadowspawn:
Thanks. I found a somewhat obscure command to tell whether OS is 32-bit or 64-bit: `getconf LONG_BIT`.

Given there is no build of node for `armv8l`, getting a false positive for the possibly rare case of 32-bit armv8, might just got with this PR without an extra test.

shadowspawn:
Released in v6.4.0

Issue #617, opened 2020-04-16
Support Node Support Aliases


shadowspawn:
I am opening this as a parallel to https://github.com/nvm-sh/nvm/issues/2170 , rather than gate-crashing the `nvm` issue.

FYI: @wesleytodd

The Node Aliases are: https://github.com/nodejs/package-maintenance/issues/236

In general, `n` resolves an alias to a single resolved version. There are not any existing commands in `n`where an alias resolves to multiple versions across major version number or release streams.

The Node Aliases which specify a single version are `lts_latest` and `current` (which are currently available in `n` as `lts` and `current`). I see those as being relevant to a node version manager, and the version can be resolved using just the information in <https://nodejs.org/dist/index.tab>.

The Node aliases which cover multiple streams are `all`, `lts`, `active`, `lts_active`. 
Is there any value in supporting these aliases in some way when `n` does not have multiple version behaviour? My feeling is that it would be inconsistent as `n run all` would effectively be identical to `n run current`, and these aliases seem more relevant in a CI setting where you want test run across a matrix of target versions.

wesleytodd:
> The Node aliases which cover multiple streams are all, lts, active, lts_active.
Is there any value in supporting these aliases in some way when n does not have multiple version behaviour?

So the way I see that working in both `nvm` and `n` is that it would compare semver and pick the highest version at the time.  If you look at the implementation in `@pkgjs/nv` it does sort by "semverness" so in the tooling I have it just plucks the first item.  If we could have the same behavior across these tools I think that would be great!

One thing brought up on `nvm` was that we might want the dates in the `.tab` or `.json` files produced by the build team.  I think this would be a great addition and if both `nvm` and `n` came to the table on this I don't know why it would be opposed.

shadowspawn:
So if I am picking the highest version at the time, are these mappings correct?

- `all` => `current`
- `lts` => `lts_latest`
- `active` => `current`
- `lts_active` => `lts_latest`

(i.e. no dates consulted, newest version is by definition active.)

shadowspawn:
A more formal list in a draft document: https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/PACKAGE-SUPPORT.md#node-name-space

shadowspawn:
I have refreshed myself on all the links and threads I could find. The closest I can find to a canonical reference is https://github.com/nodejs/package-maintenance/blob/master/docs/drafts/PACKAGE-SUPPORT.md#node-name-space
[...]

PR #619, opened 2020-04-26
Add node support aliases

shadowspawn:
# Pull Request

## Problem

Node are defining some aliases to describe the versions of node a package supports (and/or is tested against), and would like to standardise their use in their own documentation, and encourage support in other tooling.

See #617 for links and further information.

## Solution

The EOL information is not available in `index.tab` yet, so simply map the support aliases onto either `lts` or `current` from previous `n` usage.

In the future, enhancing `ls-remote` to list all the matching versions would be a nice feature and add some value to their support in `n` beyond simply allowing their use.

## ChangeLog

tba

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Execute `n` on its own to view your downloaded versions, and install the selecte
 
     Use up/down arrow keys to select a version, return key to install, d to delete, q to quit
 
-(You can also use `j` and `k` to navigate up or down without using arrows.)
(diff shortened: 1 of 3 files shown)


shadowspawn:
The document "Baseline practice - Document support levels" is just out of draft:

https://github.com/nodejs/package-maintenance/blob/master/docs/PACKAGE-SUPPORT.md#support-target

shadowspawn:
Shipped in `n` v6.6.0

PR #627, opened 2020-07-15
Use https instead of http

DanielRuf:
# Pull Request

## Problem

On npmjs I saw that there is mixed content for this package. This is due to the gif loaded over http. This lead me to some more places where http instead of https is used. This PR fixes this.

## Solution

Use https instead of http.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# EditorConfig is awesome: http://EditorConfig.org
+# EditorConfig is awesome: https://EditorConfig.org
 
 # top-most EditorConfig file
 root = true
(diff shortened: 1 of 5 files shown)


DanielRuf:
Hm, seems there is no CI for this repo. Maybe another useful addition in the future.

shadowspawn:
> Hm, seems there is no CI for this repo. Maybe another useful addition in the future.

One day... The current tests thrash downloads from nodejs, and so I run them locally with a manually launched proxy to reduce network loads.

DanielRuf:
> One day... The current tests thrash downloads from nodejs, and so I run them locally with a manually launched proxy to reduce network loads.

The CI setup could use caches and maybe a verdaccio instance may be helpful in some cases. Just some ideas.

Issue #636, opened 2020-11-24
`auto` stops at first `package.json` even without `engines.node`


mhassan1:
# Bug Report

## Summary

`n exec auto` stops at the first `package.json` it sees, even if that `package.json` does not contain `engines.node`

## Steps to Reproduce

1. Create directories `root` and `root/pkg`
2. Create a file `root/.nvmrc` containing `14`
3. Create a file `root/pkg/package.json` containing `{}`
4. In `root/pkg`, run `n exec auto node -v`

## Expected Behaviour

```
       found : /path/to/root/.nvmrc
        read : 14
v14.15.1
```

## Actual Behaviour

```
       found : /path/to/root/pkg/package.json
        read :
v15.3.0
```

## Configuration Details

```bash
$ n --version
6.7.0

$ command -v node
/Users/me/.n/bin/node

$ node -p process.platform
darwin
```

shadowspawn:
(Nice description, thanks.)

That is the intended behaviour. `n` is looking for a file to consult, and stops searching at the first file it finds. This is similar to the behaviour of `npm` itself looking for the root of the package from the current directory.

- https://docs.npmjs.com/cli/v6/configuring-npm/folders#more-information

mhassan1:
I have a monorepo with a structure like this:
```
.nvmrc
package.json
packages/
  package1/
    package.json
  package2/
    package.json
```
None of the `package.json` files have `engines.node` specified in them.

If I run `n exec auto node index.js` from within `packages/package1`, I expect it to use the version specified in `.nvmrc`. Since I haven't specified `engines.node` in the `packages/package1/package.json`, I wouldn't expect that to mean "use latest". At least for my use case, this behavior makes `auto` less useful, since I'd like to control the node version at the root without updating `engines.node` in every `package.json` file.

I think my bias here is that I am looking to get similar behavior to what I get with `nvm`, where it looks for the nearest `.nvmrc` file. Perhaps there's a feature request here where we can tweak `auto` to do the same directory traversal but only look for some file types. If you think that idea has potential, I can open a separate issue.

shadowspawn:
With that setup and bias, I see why you hoped it would work that way. 😄 

I did wonder about what to do with missing `node.engines` when I wrote this feature. Installing `current` is loosely consistent with absence of `node.engines` but perhaps not particularly useful in the absence of any constraint. I didn't want to throw an error.

[...]

PR #637, opened 2020-11-25
Fix auto support to ignore package.json when no engines found

mhassan1:
# Pull Request

## Problem

`auto` stops at the first `package.json` it sees, even if that `package.json` does not contain `engines.node` (https://github.com/tj/n/issues/636)

## Solution

If the `package.json` does not contain `engines.node`, continue looking for other `auto` files

## ChangeLog

> `auto`: Ignore `package.json` when it does not contain `engines.node`

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -949,12 +949,17 @@ function display_file_node_version() {
 
 function display_package_engine_version() {
   local filepath="$1"
-  trace_log "found" "${filepath}"
   command -v node &> /dev/null || abort "an active version of node is required to read 'engines' from package.json"
   local range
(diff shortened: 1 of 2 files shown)


shadowspawn:
On way of thinking about this is whether the current package is a bounding context for finding an appropriate node version.

A couple of scenarios where scanning further would be useful for non-package guidance:
- package is in a mono repo, and a control file is at the root of the mono repo
- package is under users home directory, and a control file is in the home directory

shadowspawn:
Skipping an empty `node.engines` field was worth considering, but #639 points out that a `node.engines` field of `*` leads to the same problems, so not a solution to the underlying issue for a mono-repo setup with a version control file higher up the folder hierarchy

Likely to take another approach. Later issues: #639 #640

Issue #642, opened 2020-12-14
Error: download preflight failed


al3x-huang:
# Bug Report

## Summary

When running `n <version>` I get this error message:
```
$ n 10

/usr/local/bin/n: line 194: [[: v10: syntax error: operand expected (error token is "v10")

  installing : node-vv10.23.0
curl: (3) [globbing] bad range in column 28

  Error: download preflight failed for 'v10.23.0' (https://nodejs.org/dist/vv10.23.0/node-vv10.23.0-darwin-x64.tar.xz)
```

it seems like the URL is incorrect: `vv10.23.0` in the above example instead of `v10.23.0`

## Steps to Reproduce

`n 10` or `n <version>` or `n v10`

## Expected Behaviour

Install node correctly

## Actual Behaviour

Throws error

## Other Information

`n doctor` runs fine
`n -V` is 6.8.0
`n 10.23.0` is the only command that runs/installs

## Configuration Details

```bash
$ n --version
6.8.0

$ command -v node
/usr/local/bin/node

$ node -p process.platform
darwin
```

al3x-huang:
It appears after looking through `/usr/local/bin/n` there is a problem with the version string and how it's generated.
```
install() {
  [[ -z "$1" ]] && abort "version required"
  local version
  #version="$(display_latest_resolved_version "$1")" || return 2
  version="10.23.0"
  ...
```

hardcoding `version` without the `v` prefix produces the expected behavior. Perhaps someone should take a look at `display_latest_resolved_version` and generate the version string without the `v` prefix?

shadowspawn:
I am not seeing the same problem, so something different between our setups. There is a command which is used by the units tests which will let us conveniently compare calling display_latest_resolved_version. What do you see when you run this:

```sh
$ n N_TEST_DISPLAY_LATEST_RESOLVED_VERSION 10
10.23.0
```

al3x-huang:
```
$ n N_TEST_DISPLAY_LATEST_RESOLVED_VERSION 10
v10.23.0
```

al3x-huang:
Curiously, 
```
$ n N_TEST_DISPLAY_LATEST_RESOLVED_VERSION 10.23.0
10.23.0
```

shadowspawn:
There is code which is meant to be dropping the `v` from the string using grep and an extended regular expression. What do you see with this command:

```sh
 $ echo "v10.23.0" | grep -E -o '[^v].*'
10.23.0
```

[...]

PR #643, opened 2020-12-19
Fix wrong version string match for colorized grep

al3x-huang:
* Add a wrapper function for grep to ensure consistency across different
  aliases and options by forcing it to use builtin grep without GREP_OPTIONS

# Pull Request

## Problem #642 

When running `n <version>` I got this error message:
```
$ n 10

/usr/local/bin/n: line 194: [[: v10: syntax error: operand expected (error token is "v10")

  installing : node-vv10.23.0
curl: (3) [globbing] bad range in column 28

  Error: download preflight failed for 'v10.23.0' (https://nodejs.org/dist/vv10.23.0/node-vv10.23.0-darwin-x64.tar.xz)
```

it seemed like the URL is incorrect: `vv10.23.0` in the above example instead of `v10.23.0`

The problem turned out to be because I had enabled colorized grep a long time ago in `.bash_profile` as it is, in most use cases, helpful for me to highlight matches. 
```
# color grep
export GREP_OPTIONS='--color=always'
```
However, some regex patterns in `n` matched on these ANSI color codes produced by colorized grep:
```
^[[01;31m^[[Kv10.^[[m^[[K23.0   2020-10-27  aix-ppc64,headers,linux-arm64,linux-armv6l,linux-armv7l,linux-ppc64le,linux-s390x,linux-x64,osx-x64-pkg,^[[01;31m^[[Kosx-x64-tar^[[m^[[K,src,sunos-x64,win-x64-7z,win-x64-exe,win-x64-msi,win-x64-zip,win-x86-7z,win-x86-exe,win-x86-msi,win-x86-zip    6.14.8  6.8.275.32  1.34.2  1.2.11  1.1.1g  64  Dubnium -
```
and produced an incorrect version string

## Solution

Create and use a wrapper function for grep that ensures we use the built-in `grep` without any additional `GREP_OPTIONS`.

```
function n_grep() {
  GREP_OPTIONS='' command grep "$@"
}
```

## ChangeLog

Fix bug caused by colorized grep output

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -40,6 +40,15 @@ function echo_red() {
 }
 
 #
+# Synopsis: grep [args] [pattern]
+# Grep wrapper to ensure consistent grep options and circumvent aliases
+#
+
+function n_grep() {
+  GREP_OPTIONS='' command grep "$@"
+}
+
+#
 # Setup and state
 #
 
@@ -383,7 +392,7 @@ err_no_installed_print_help() {
 #
 
 function next_version_installed() {
-  display_cache_versions | grep "$1" -A 1 | tail -n 1
+  display_cache_versions | n_grep "$1" -A 1 | tail -n 1
 }
 
 #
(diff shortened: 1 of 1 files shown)


shadowspawn:
LGTM, thanks. I'll spin up in editor and run the tests before approving.

shadowspawn:
Released in v7.0.0

Issue #640, opened 2020-11-27
Poll: move package.json engine support out of auto to a new command?


shadowspawn:
The current behaviour with `n auto` is to stop scanning at the first candidate file which includes `package.json`. This prevents scanning outside a package for a control file, say `.node-version` at the root of a mono repo or in the users home directory.

The purpose of `engine` in `package.json` is rather different than a version control file, and processing `package.json` uses extra dependencies.

One approach would be to use different commands for finding the various node version files like `.node-version`, and for finding node version from `package.json`. The new command could be called `n engine` or `n package`.

Related: #618 #636 #639

Show your preference with a reaction:
👍 move `package.json` support out of `auto` to a new command
😐 single command is nice, but two commands would be fine
👎 don't change, keep package.json support in `auto`

mhassan1:
@shadowspawn when might you expect to close the poll?

shadowspawn:
At this stage, 5 replies or a month, whichever comes first. Eagerly waiting? 😄

shadowspawn:
(Note to self: if `auto` fails to find a version, error could suggest `engine` so clue about change in behaviour.)

shadowspawn:
I am having second thoughts about splitting out engine as a I like `n auto` for getting `n` to do something sensible.

Wondering if first search for the version control files all the way up hierarchy, and only then look for `package.json`.

i.e.
1) search up hierarchy for `node-version`, `npmrc` et al
2) fallback look for `package.json`

This works same as before for simple cases of:
- only version control file
- only package.json

While supporting the mono-repo case of a higher level version control file that is more appropriate for picking a node version than the local package.json file.

mhassan1:
That works for me, and I think it's a sensible compromise.

Issue #639, opened 2020-11-25
Feature Request: Ability to configure `auto` file types


mhassan1:
# Problem

## Short Version

It would be useful to be able to configure which file types are crawled by `auto`

## Long Version

Support for `.nvmrc` files in the `auto` feature is awesome! Our developers use `nvm` for managing node versions in different projects, and we are using `n` for managing node versions in our CI/CD. It's cool to be able to use both tools with only one configuration file.

However, the fact that a `package.json` file (especially one without a `engines.node` setting) interrupts crawling for `.nvmrc` files makes this feature less useful in a monorepo setting. Here is one use case: https://github.com/tj/n/issues/636#issuecomment-733433947

Here is a list of cases where crawling all the file types is undesirable:
* A `package.json` file with no `engines.node` interrupts crawling for other file types
* A `package.json` file with a very permissive `engines.node` setting (e.g. `"*"`) interrupts crawling for other file types (since `engines.node` may serve a different purpose than `.nvmrc` in a developer's home directory, for example)
* A `package.json` file is found but there is no `node` or `npx` binary available yet (currently, this is an error, which penalizes the user for using `auto` even if they aren't using `engines.node`)

Some potential solutions:
* Add additional presets for single file types, e.g. `n .nvmrc` or `n auto-.nvmrc` or `n auto[.nvmrc]`
* Add support for specifying a list of file types, e.g. `n auto[.nvmrc,package.json]`
* Add a flag to skip `package.json`, similar to https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-ignore-engines (this is a stretch)

## Configuration Details

```bash
$ n --version
6.7.0

$ command -v node
/Users/me/.n/bin/node

$ node -p process.platform
darwin
```

shadowspawn:
Are all your use cases solved by removing the scanning of `package.json`?

I almost did it separately as `n engine` originally, but went for `auto` as a single solution. 

I am interested in whether the people who want and use the `package.json` support mix it with other version sources (like `.nvmrc`), or whether having separate commands would be fine and more flexible. I'll open another issue to poll for feedback.

shadowspawn:
Of the three potential solutions suggested, I currently prefer a flag to ignore engine/package. Scanning `package.json` is the outlier in terms of function and requirements.

shadowspawn:
Another approach, make `package.json` a file of last resort. Only look for `package.json` after searching up hierarchy for the other (better) version control files. See #644

[...]

PR #644, opened 2020-12-20
WIP: Scan package.json last

shadowspawn:
# Pull Request

## Problem

Including `package.json` in `auto` blocks the use of the more explicit version control files further up the folder hierarchy, particularly in mono-repo layout. The `package.json` is not likely to as good a source of a specific version to install as the other version control files.

Related: #636 #639 #640 #637

## Solution

- `auto` scans for `package.json` only after looking for other (better) files up hierarchy.
- add `engine`
- rework tests to remove misuse of BATS `run` and make tests more maintainable

### Changed

- `auto` label now scans for `package.json` only if it can not find a version control file (#644)

### Added

- `engine` label to look for `engines.node` in `package.json` (as used by`auto`)


shadowspawn:
Release in v7.0.0

Issue #641, opened 2020-11-27
"no version found" error on Mac with M1 chip


shadowspawn:
Using `n current` or `n 8` etc on a Mac with the M1 chip and an older version of `n` will currently display an error that "no version found", because there are not yet native versions of node available for the new architecture.

You can download and run node built for the Intel x64 architecture. The installed node will run using Rosetta 2. e.g.

```sh
n --arch x64 lts
```

Another approach is to launch Terminal itself using Rosetta, so that the terminal environment assumes x64.

The `n` behaviour on Apple M1 will be reviewed when nodejs decide their plans for support:

- https://github.com/nodejs/build/issues/2474 
- https://github.com/nodejs/TSC/issues/886

soffes:
`n latest` on an M1 shows `Error: no version found for 'latest'`

shadowspawn:
Your situation is what this issue is covering @soffes . For now, you can do installs by specifying the architecture.

shadowspawn:
I have been wondering about releasing a version of `n` which defaults to x64 for now, and fine tune it after nodejs start releasing native versions. 

Readers, let me know if this would be useful with a 👍 or comment.

shadowspawn:
Released `n` v6.8.0 which defaults to x64 on a Mac with the M1 chip.

(The native architecture is `arm64` if you need to specify it to try a new release before `n` catches up. `n` does not separate the downloads by architecture, so if you are changing architectures for a specific version then delete the old download before installing again with the desired architecture.)

mattdamon108:
@shadowspawn Recently, I got a M1 air, then I installed the latest version of n. But, it installed the x64 arch node v14 lts after the normal initial installation.

Can you help me to install the arm64 node version with n?

shadowspawn:
@mattdamon108 There are not yet any prebuilt arm64 binaries available for Mac:

https://nodejs.org/dist/latest/

mattdamon108:
@shadowspawn oh! thank you for clarification!

parndt:
Looks like there are now, at least version 15.7.0 has a prebuilt arm64 for darwin - https://nodejs.org/dist/v15.7.0/node-v15.7.0-darwin-arm64.tar.xz
```
 ✗ n 15
  installing : node-v15.7.0
       mkdir : /Users/parndt/n/n/versions/node/15.7.0
       fetch : https://nodejs.org/dist/v15.7.0/node-v15.7.0-darwin-x64.tar.xz
[...]

PR #663, opened 2021-04-16
Build from source on apple silicon, as long as version is above 15

thtliife:
# Pull Request

## Problem

No arm64 binary found on m1 apple silicon
issue: #641

## Solution

If no tarball found on Darwin os with arm64 arch, falls back to building from source tarball, as long as version is above 15.
I have successfully built the following versions:
```bash
node/12.22.1
node/13.14.0
node/14.5.0
node/14.16.1
node/15.0.0
node/15.12.0
node/15.14.0
```
but according to nodejs issue [#37309, Support is very unlikely to be back ported to 14.x or before due to requiring a SemVer Major V8 update](https://github.com/nodejs/node/issues/37309)

## ChangeLog

Fallback to building from source if no prebuilt binary found

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -698,6 +698,66 @@ activate() {
 }
 
 #
+# Build <version> from source
+#
+build() {
+  [[ -z "$1" ]] && abort "version required"
+  local version
+  get_latest_resolved_version "$1" || return 2
+  version="${g_target_node}"
+  [[ -n "${version}" ]] || abort "no version found for '$1'"
+  update_mirror_settings_for_version "$1"
+  update_xz_settings_for_version "${version}"
+
+  local dir="${CACHE_DIR}/${g_mirror_folder_name}/${version}"
+
+  if test "$N_USE_XZ" = "true"; then
+    local tarflag="-Jx"
+  else
+    local tarflag="-zx"
+  fi
+
+  if test -d "$dir"; then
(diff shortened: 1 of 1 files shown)


shadowspawn:
Related: the nightly builds currently have an arm64 archive, although the index has not yet been updated.

https://nodejs.org/download/nightly/v16.0.0-nightly202104148d9d8236b7/

thtliife:
> Related: the nightly builds currently have an arm64 archive, although the index has not yet been updated.
> 
> https://nodejs.org/download/nightly/v16.0.0-nightly202104148d9d8236b7/

You're right...
I just tried "n nightly" and got the following:
n nightly terminal output

running n after that results in:
n menu

So maybe this pull request is not gonna be required...?

shadowspawn:
Building from source would be a big change and potential support burden with people asking questions about build or even runtime problems. For example, the Node.js building page is rather long.

- https://github.com/nodejs/node/blob/master/BUILDING.md

I'm happy sticking with the pre-build binaries, which are just becoming available (hurrah!). I just saw a tweet this morning that Node 16 coming and featuring Apple silicon binaries:

- https://twitter.com/BethGriggs_/status/1383010606660325377

rockyedge:
Rejected quit apple immediately signed tollew

PR #664, opened 2021-04-17
Add support for new Node.js builds for Apple silicon

shadowspawn:
# Pull Request

## Problem

Node.js is available natively on Apple silicon from v16, but not before.

`n` is currently switching the architecture from arm64 to x64 on Darwin to allow installing older versions to run using Rosetta.

Related: #641

## Solution

On Macs with Apple silicon...

- version resolution and remote listing include both the x64 and arm64 targets
- downloads for Node.js 16 and higher use arm64
- downloads for Node.js below 16 use x64
- specifying an explicit architecture using `--arch` shows and downloads only the matching architecture

I considered switching architecture if arm64 download missing, but decided simpler code and story switching based on the Node.js version.

## ChangeLog

- on Macs with Apple silicon, use native arm64 for Node.js versions for version 16 and higher

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- markdownlint-disable MD024 -->
 
-## [Unreleased] (date goes here)
+## [7.2.0] (2021-04-18)
+
+### Added
+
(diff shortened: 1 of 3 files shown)


shadowspawn:
This PR assumes `index.tab` lists the arm64 as an available download, which isn't happening quite yet.

See https://github.com/nodejs/nodejs-dist-indexer/issues/14

PR #669, opened 2021-05-03
Add Emacs keyboard shortcuts for menu navigation

aidin36:
# Pull Request

## Problem

By default, both Bash and Zsh accept Emacs default short-keys, including up and down keys (`ctrl-p` and `ctrl-n`). It's a good idea to support those short-keys to make it easier for people who used to use those short-keys.

## Solution

I added both `ctrl-p` and `ctrl-n` keys to the switch-case.

## ChangeLog

Add support for Emacs up and down keys (`ctrl-p` and `ctrl-n`)

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -527,6 +527,8 @@ menu_select_cache_versions() {
   ESCAPE_SEQ=$'\033'
   UP=$'A'
   DOWN=$'B'
+  CTRLP=$'\020'
+  CTRLN=$'\016'
 
   while true; do
     read -rsn 1 key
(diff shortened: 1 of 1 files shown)


shadowspawn:
An example of `ctrl+n` and `ctrl-p` listed as control key combinations: https://en.wikipedia.org/wiki/Control_key#Table_of_examples

shadowspawn:
I have not seen this being asked for, but the ket sequence used in multiple contexts and should not cause problems.

aidin36:
Thanks for merging it! :slightly_smiling_face:
It makes its behaviour more consistent with *nix terminals.

shadowspawn:
Released in v7.3.0

Issue #674, opened 2021-05-31
ls-remote should support 'engine'


mykhalov:
# Bug Report

`ls-remote` should support 'engine' option as the version as `install` does.

## Steps to Reproduce

1. Have a package.json with `"engine": { "node": "12" }`
2. Run `n ls-remote engine`

## Expected Behaviour

An output identical to `n ls-remote 12`

## Actual Behaviour

```
Error: invalid version 'engine'
```

## Configuration Details

```bash
$ n --version
7.2.2

$ command -v node
12.8.0

$ node -p process.platform
darwin
```

shadowspawn:
Why do want to run `n ls-remote engine`? Are you running it yourself to see what version would be installed, or all matching versions, or perhaps wanting to use it in a script to find resolved version?

(I decided not to implement `ls-remote` at the time I added `auto` and `engine` as didn't seem a good fit and was extra work, so interested what you want to use it for.)

mykhalov:
My use case is to print what exact version will be installed:

```
% n ls-remote 12 --all | head -1
12.22.1
```

Not being able to use `version` here feels like a bug and isn't mention in the docs:

```
% n --help

Usage: n [options] [COMMAND] [args]

Commands:

[...]
  n install <version>            Install Node.js <version> (downloading if necessary)
[...]
  n ls-remote [version]          Output matching versions available for download
[...]

Versions:
[...]
    engine            Read version from package.json
[...]
```

shadowspawn:
Released in v7.3.0

PR #675, opened 2021-06-05
engine and auto improvements

shadowspawn:
# Pull Request

## Problem

The `engine` and `auto` labels do not work for `ls-remote` (and are not documented as being special in this way).

Issues: #674

## Solution

Refactor the handling of `engine` and `auto` so also works for `ls-remote`. 

To allow logging from "display" routines the diagnostic output during the file scanning for `engine` and `auto` was moved from stdout to stderr, and can now be suppressed with `--quiet`.

This allows code like:

```
desired_version=$(N_MAX_REMOTE_MATCHES=1 n --quiet ls-remote auto)
```

## ChangeLog

- add: `ls-remote` supports `engine` and `auto` labels
- add: reduce `engine` and `auto` logging with `--quiet`
- changed: diagnostic logging during processing of `engine` and `auto` written to stderr rather than stdout

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -11,12 +11,15 @@ log() {
 }
 
 #
-# diag <type> <msg>
-# Same as log, but diagnostic to stderr rather than stdout.
+# verbose_log <type> <msg>
+# Can suppress with --quiet.
+# Like log but to stderr rather than stdout, so can also be used from "display" routines.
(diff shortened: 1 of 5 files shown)


shadowspawn:
Released in v7.3.0

PR #681, opened 2021-07-25
feat(docs): add shell syntax highlighting

mesqueeb:
# Pull Request
## Problem

No shell syntax highlighting in readme.

## Solution

Add shell syntax highlighting in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,31 +35,39 @@ It is written as a BASH script but does not require you to use BASH as your comm
 
 Since you probably already have `node`, the easiest way to install `n` is through `npm`:
 
-    npm install -g n
+```bash
+npm install -g n
+```
 
 Once installed, `n` caches Node.js versions in subdirectory `n/versions` of the directory specified in environment variable `N_PREFIX`, which defaults to `/usr/local`; and the _active_ Node.js version is installed under `N_PREFIX`.
 
 To avoid requiring `sudo` for `n` and `npm` global installs, it is suggested you either install to your home directory using `N_PREFIX`, or take ownership of the system directories:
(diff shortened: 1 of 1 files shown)


shadowspawn:
I had used indentation code code blocks mainly because it was cleaner for editing, but worth considering changing as the viewing matters more than the editing.

The syntax highlighting I see in README for https://github.com/mesqueeb/n/tree/patch-2 is a bit underwhelming. I see `export` in colour in the examples near the end, but also false-positive highlights for `select` and `return` and `exec` and `test` in other code blocks. I am not seeing a lot of value in the highlighting? (I assume the syntax highlighting does not vary much by browser. I did check with both Safari and Chrome.)

I did some reading and see `ShellSession` (aka console) suggested for command-line interactions, as opposed to actual shell scripts.

- https://stackoverflow.com/a/49004070/1082434
- https://github.com/github/linguist/blob/master/vendor/README.md

shadowspawn:
I tried `ShellSession` and not sure what the rules are, but mainly see a colour difference between lines starting with $ and those without. I have been following a `markdownlint` suggestion to only use $ when also showing the output, so simply using `ShellSession` without changing the code blocks does not work very well.

https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md014

mesqueeb:
@shadowspawn I really enjoy that comments behind `#` are light grey, so it's easy to differentiate comments from commands.
the rest I don't mind so much.

shadowspawn:
I like the comments too. If you just change that one, I'll merge.

PR #534, opened 2018-10-28
Mention brew install n

joelpurra:
- The popular macOS package manager `brew` has a formula for `n`, maintained by the Homebrew community.
- Does not require installing `node`/`npm` first.
- The `n` formula was installed 1401 times in the past 30 days (as of 2018-10-28T09:01:04Z).
- This commit adds the installation method below the `npm` and `make` methods, but above `n-install`, in `README.md`.

```shell
brew install n
```

See

- https://brew.sh/
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/n.rb
- https://formulae.brew.sh/analytics/install/30d/
- https://formulae.brew.sh/api/analytics/install/homebrew-core/30d.json

PR #685, opened 2021-09-05
Add support for opt-in download to exec and run

### LINKED ISSUE #647: Suggestion: n run should download first
I think it would be handy if `n run` could download the version first before executing, instead of relying on the cached binary being present. This would be handy if you are trying to use `n run engine` to use the matching version from package.json (but may not always exist in cache). 

Why can't `n run` act similarly to how `n use` works to download the version first before setting it as default binary?

shadowspawn:
# Pull Request

## Problem

It is inconvenient to use `n run` and `n exec` in CI situations where the version may not have already been downloaded. 

Workarounds are to have some extra script steps or setup to always install, or conditionally install. But this may be difficult to arrange in an npm run script, for example.

Resolves #647

## Solution

Allow use of the existing `-d, --download` flag with `run` and `exec` to download the target version, if necessary.

This is a slightly different use of the flag than with an install, but I didn't come up with a better flag and I think the usage is clear enough in both cases.

```
n --download lts
n run --download latest myScript.js
```

Example usage:

```
% node --version
v10.24.1
% n run lts --version

  Error: 'lts' (14.17.6) not in downloads cache

% n run -d lts --version
  installing : node-v14.17.6
       mkdir : /usr/local/n/versions/node/14.17.6
       fetch : https://nodejs.org/dist/v14.17.6/node-v14.17.6-darwin-x64.tar.xz
v14.17.6
% node --version
v10.24.1
```

(A side-effect of the implementation is that `--download` is also supported for `which`!)

## ChangeLog

- add support for `--download` option to `run` and `exec` to download the target version, if necessary

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -115,9 +115,9 @@ if [ -n "$HTTP_USER" ];then
   if [ -z "$HTTP_PASSWORD" ]; then
     abort "Must specify HTTP_PASSWORD when supplying HTTP_USER"
   fi
-  CURL_OPTIONS+=("-u $HTTP_USER:$HTTP_PASSWORD")
-  WGET_OPTIONS+=("--http-password=$HTTP_PASSWORD"
-                "--http-user=$HTTP_USER")
+  CURL_OPTIONS+=( "-u $HTTP_USER:$HTTP_PASSWORD" )
+  WGET_OPTIONS+=( "--http-password=$HTTP_PASSWORD"
+                "--http-user=$HTTP_USER" )
 elif [ -n "$HTTP_PASSWORD" ]; then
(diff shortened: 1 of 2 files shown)


jaymefSO:
I noticed that brew still has `n` forumla for version 7.3.1. Who maintains that? Can we get the latest version bumped on brew?

shadowspawn:
The third-party installers are maintained by others, and I don't know how quickly they pick up new versions. To be fair, it has not been long since 7.4 was released.

I did a little digging:
- https://docs.brew.sh/Updating-Software-in-Homebrew
- HomeBrew added to README: #534

PR #689, opened 2021-09-23
README - clearer show what directories are created under N_PREFIX

paulvi:
# Pull Request

## Problem

I used `~/opt` as value for N_PREFIX and got a bit of mess there.

N_PREFIX would better be named N_HOME or explained this way.

## Solution

show what directories are created, so that would be no surprises.

I now personally use `~/opt/nodejs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ If you already have Node.js installed, an easy way to install `n` is using `npm`
 
     npm install -g n
 
-Once installed, `n` caches Node.js versions in subdirectory `n/versions` of the directory specified in environment variable `N_PREFIX`, which defaults to `/usr/local`; and the _active_ Node.js version is installed under `N_PREFIX`.
+Once installed, `n` caches Node.js versions in subdirectory `n/versions` of the directory specified in environment variable `N_PREFIX`, which defaults to `/usr/local`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,9 @@ If you already have Node.js installed, an easy way to install `n` is using `npm`
 
     npm install -g n
 
-Once installed, `n` caches Node.js versions in subdirectory `n/versions` of the directory specified in environment variable `N_PREFIX`, which defaults to `/usr/local`; and the _active_ Node.js version is installed under `N_PREFIX`.
+Once installed, `n` caches Node.js versions in subdirectory `n/versions` of the directory specified in environment variable `N_PREFIX`, which defaults to `/usr/local`.
+
+The _active_ Node.js version is installed under `N_PREFIX`. (This creates subdirectories `bin`, `include`, `lib` and `share`.)
```
shadowspawn, line 42: Perhaps "uses" rather than "creates"? In the default location the subdirectories already exist.

paulvi: In my current understanding, installing version does creates new folders under N_PREFIX location

And if N_PREFIX points to some system folder, then installing will fail without `sudo`

shadowspawn:
Merged onto `develop` branch by hand.

Issue #706, opened 2022-01-06
dash does not refresh its hash table when PATH is re-declared


agarzola:
The following line recommends re-declaring `PATH` to ensure the new location of Node.js is loaded in the current session.

https://github.com/tj/n/blob/d569e331820399ae5224b640f1d556c64a87729d/bin/n#L744

However, in some environments that might not suffice. I ran into this issue on Platform.sh, which uses [`dash`](https://man7.org/linux/man-pages/man1/dash.1.html). Interestingly enough, after installing Node with `n auto` (our project has an `.nvmrc` file specifying v14), running `which node` points to the location where we told `n` to install Node: `/app/.global/bin/node`. However, running `node --version` yielded v12. Simply invoking `export PATH="$PATH"` did not solve this issue.

As it turns out, dash’s implementation of the `hash` built-in command does not refresh the hash table where executables are stored for easy access. The fix suggested to us by Platform.sh support was to invoke `hash -r` instead. This worked: after invoking that, `node --version` yielded the version number we expect (the one installed via `n`).

I will submit a PR shortly that updates the line linked above to include a note about how to solve this issue in dash.

shadowspawn:
I remember considering `hash -r` vs `PATH=$PATH` but can't find any notes I made at the time.

Related issues mentioned in the CHANGELOG: advice to reset command hash when node location changes (#170 #381 #451 #588)

shadowspawn:
More documentation:
- https://hyperpolyglot.org/unix-shells#cmd-resolution
- https://man7.org/linux/man-pages/man1/dash.1.html
- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/hash.html

The Open Group docs for hash say this, which is probably why I went for `PATH` over `hash` originally:

> The effects of hash -r can also be achieved portably by resetting the value of PATH ; in the simplest form, this can be:
>
> PATH="$PATH"

agarzola:
That is true of Bash’s implementation of `hash`. Alas, that appears to not be the case in Dash’s implementation.

shadowspawn:
Yes "portably" is not universally. 🙊

shadowspawn:
After some experimentation...

```
bash: PATH=$PATH or hash -r
fish: no action required
ksh: hash -r
tcsh: rehash
zsh: PATH=$PATH or hash -r
dash: hash -r
csh: rehash
ash: hash -r
```

agarzola:
Oh, wow. So how about I modify my PR thusly:

[...]

PR #707, opened 2022-01-06
GH-706 Update location change message

agarzola:
# Pull Request

This pull request updates the message printed for a user when installing a new version of Node.js via `n` results in a location change for the binary.

Closes GH-706.

## Problem

One of the recommended solutions does not work in Dash:
```
Note: the node command changed location and the old location may be remembered in your current shell.
         old : /usr/bin/node
         new : /app/.global/bin/node
To reset the command location hash either start a new shell, or execute PATH="$PATH"
```

See GH-706.

## Solution

Update the message to suggest an alternative if using Dash:
```
Note: the node command changed location and the old location may be remembered in your current shell.
         old : /usr/bin/node
         new : /app/.global/bin/node
If "node --version" shows the old version then start a new shell, or reset the location hash with:
hash -r  (for bash, zsh, ash, dash, and ksh)
rehash   (for csh and tcsh)
```

## ChangeLog

Update location change message with recommendation for Dash.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -741,7 +741,7 @@ activate() {
       log old "${original_node}"
       log new "${active_node}"
       # shellcheck disable=SC2016
-      printf 'To reset the command location hash either start a new shell, or execute PATH="$PATH"\n'
+      printf 'If "node --version" shows the old version then start a new shell, or reset the location hash with:\nhash -r  (for bash, zsh, ash, dash, and ksh)\nrehash   (for csh and tcsh)'
     fi
   fi
 }
--- a/bin/n
+++ b/bin/n
@@ -740,7 +740,6 @@ activate() {
       printf '\nNote: the node command changed location and the old location may be remembered in your current shell.\n'
       log old "${original_node}"
       log new "${active_node}"
-      # shellcheck disable=SC2016
       printf 'If "node --version" shows the old version then start a new shell, or reset the location hash with:\nhash -r  (for bash, zsh, ash, dash, and ksh)\nrehash   (for csh and tcsh)'
     fi
   fi
--- a/bin/n
+++ b/bin/n
@@ -740,7 +740,7 @@ activate() {
       printf '\nNote: the node command changed location and the old location may be remembered in your current shell.\n'
       log old "${original_node}"
       log new "${active_node}"
-      printf 'If "node --version" shows the old version then start a new shell, or reset the location hash with:\nhash -r  (for bash, zsh, ash, dash, and ksh)\nrehash   (for csh and tcsh)'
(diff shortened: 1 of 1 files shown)


bin/n
```diff
@@ -740,8 +740,7 @@ activate() {
       printf '\nNote: the node command changed location and the old location may be remembered in your current shell.\n'
       log old "${original_node}"
       log new "${active_node}"
-      # shellcheck disable=SC2016
```
agarzola, line 743: @shadowspawn I noticed this shellcheck exception does not seem to apply anymore, so I removed it as well.

agarzola (commented):
This is ready for final review!

bin/n
```diff
@@ -740,8 +740,7 @@ activate() {
...
-      printf 'To reset the command location hash either start a new shell, or execute PATH="$PATH"\n'
+      printf 'If "node --version" shows the old version then start a new shell, or reset the location hash with:\nhash -r  (for bash, zsh, ash, dash, and ksh)\nrehash   (for csh and tcsh)'
```
shadowspawn, line 743: There needs to be a trailing `/n` included in the string, as printf does not add a newline.

bin/n
shadowspawn: Yes, that was due to a warning about the `$PATH` inside single quotes, which is gone.

bin/n
agarzola: Nice catch! See f7471e6.

shadowspawn:
Thanks. This situation is very confusing when a user encounters it for the first time and makes `n` seem broken, so worth trying to help as much as possible.

shadowspawn:
Example of new message in action:

```
$ export N_PREFIX=$HOME/.n
$ export PATH=$N_PREFIX/bin:$PATH
$ node --version
v16.13.1
% n 14
  installing : node-v14.18.2
       mkdir : /Users/john/.n/n/versions/node/14.18.2
       fetch : https://nodejs.org/dist/v14.18.2/node-v14.18.2-darwin-x64.tar.xz
   installed : v14.18.2 (with npm 6.14.15)

Note: the node command changed location and the old location may be remembered in your current shell.
         old : /usr/local/bin/node
         new : /Users/john/.n/bin/node
If "node --version" shows the old version then start a new shell, or reset the location hash with:
hash -r  (for bash, zsh, ash, dash, and ksh)
rehash   (for csh and tcsh)
$ node --version
v16.13.1
$ hash -r
$ node --version
v14.18.2

```

agarzola:
Thanks for letting me make this small contribution! I've so far had a delightful experience with n and cannot thank you enough for maintaining this project.

shadowspawn:
Improvements included in `n` v8.0.2

PR #708, opened 2022-01-10
Update China Mirror URL to the New Version

hc2937:
# Pull Request

https://npm.taobao.org is now redirect to https://npmmirror.com, use the new URL instead.

## Problem

Switched to the new mirror URL.

## Solution

Use the new URL.


shadowspawn:
Copy of announcement: https://segmentfault.com/a/1190000040936469/en

> As the new domain name has been officially launched, the old http://npm.taobao.org and http://registry.npm.taobao.org domain names will stop serving at 0:00 on May 31, 2022.

shadowspawn:
Thanks.

Merged to `master` branch (so visible on GitHub for people checking the home page) but not published in a release yet.

PR #717, opened 2022-03-14
Make cache directory configurable

### LINKED ISSUE #715: make cache dir configurable
currently CACHE_DIR is hardcoded to
use `$N_PREFIX/n/versions` as persistence
layer for node versions.

we use `n` in a CI context and manage
the node versions with a privileged user
while the CI user is only allowed to
switch between existing/cached versions.

currently we solved this with `$N_PREFIX` 
being owned by the CI user and 
`$N_PREFIX/n` being owned by the 
privileged user to prevent 
writing/downloading new versions during
CI jobs.

it should probably be sufficient to
make `CACHE_DIR` configurable just like
`N_PREFIX`, defaulting to the currently
hardcoded value (and also rename it to
`N_CACHE_DIR` to be unambiguous)

p.s.: yes i know, you can bypass this
restriction by setting N_PREFIX to a 
different location, but that is a 
different problem.

UiP9AV6Y:
# Pull Request

Read the cache directory from the environment, falling back to
a location under `N_PREFIX`

## Problem

fixes #715

## Solution

Initialize the cache directory variable in a similar fashion
as `N_PREFIX`


shadowspawn:
Nice work adding a test to go with the feature.

I think this would be tidier with `N_CACHE_PREFIX` matching `N_PREFIX`, and the `n/versions` getting added onto that as before. Exposing the internal `CACHE_DIR` directly has exported a directory from deeper in the `n` controlled folder, so for example can't add another folder beside "versions" in future. The internal `CACHE_DIR` can then be constructed much as before.

Simple story:
- `N_PREFIX` defaults to `/usr/local`
- `N_CACHE_PREFIX` defaults to `$N_PREFIX`

UiP9AV6Y:
i agree with the point about the lack of extendability in the future. i have implemented the changes based on your
suggestion.

shadowspawn:
Looks good. I'll read again carefully and do some local testing before merging.

PR #724, opened 2022-04-26
Chore: replace `git.io` in installation guide

SukkaW:
# Pull Request

## Problem

All links on git.io will stop redirecting after April 29, 2022: https://github.blog/changelog/2022-04-25-git-io-deprecation/

## Solution

Update the installation guide with the original URL.

## ChangeLog

- Update the installation guide with the original `n-install` script URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Or on macOS with [MacPorts](https://www.macports.org/) you can install the [n po
 
 On Linux and macOS, [n-install](https://github.com/mklement0/n-install) allows installation directly from GitHub; for instance:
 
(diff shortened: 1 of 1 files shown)


shadowspawn:
Thanks. I'll follow the UTL being used by `n-install`, which has now changed thanks to your work. Would you like to update this PR to match?

SukkaW:
> Thanks. I'll follow the UTL being used by `n-install`, which has now changed thanks to your work. Would you like to update this PR to match?

Sure. I am updating the link to the new `https://bit.ly/n-install`

shadowspawn:
Thanks for the update

PR #745, opened 2022-11-01
$< make

kaiquekandykoga:
Apparently $< is not working propertly on FreeBSD.

```
[kandy@ ~/a/n]$ make install
mkdir -p /usr/local/
cp  /usr/local/
usage: cp [-R [-H | -L | -P]] [-f | -i | -n] [-alpsvx] source_file target_file
       cp [-R [-H | -L | -P]] [-f | -i | -n] [-alpsvx] source_file ... target_directory
*** Error code 64

Stop.
make: stopped in /usr/home/kandy/a/n
```

Changing the Makefile from $< to bin/n works.

```
[kandy@ ~/a/n]$ freebsd-version
13.1-RELEASE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 PREFIX ?= /usr/local
 
 install: bin/n
-	mkdir -p $(PREFIX)/$(dir $<)
-	cp $< $(PREFIX)/$<
+	mkdir -p $(PREFIX)/$(dir bin/n)
+	cp bin/n $(PREFIX)/bin/n
 
 uninstall:
 	rm -f $(PREFIX)/bin/n
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PREFIX ?= /usr/local
 
 install: bin/n
-	mkdir -p $(PREFIX)/$(dir bin/n)
+	mkdir -p $(PREFIX)/bin
 	cp bin/n $(PREFIX)/bin/n
 
 uninstall:


Makefile
```diff
@@ -1,8 +1,8 @@
 PREFIX ?= /usr/local
 
 install: bin/n
-	mkdir -p $(PREFIX)/$(dir $<)
-	cp $< $(PREFIX)/$<
+	mkdir -p $(PREFIX)/$(dir bin/n)
```
shadowspawn, line 4: Simplify the (now) hard-coded directory name:
```suggestion
	mkdir -p $(PREFIX)/bin
```

shadowspawn (commented):
One small simplification.

Thanks for fix.

shadowspawn:
For interest, `$<` is an automatic variable in [GNU make](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html#index-automatic-variables) for "the name of the first prerequisite":



The FreeBSD man page for [make](https://www.freebsd.org/cgi/man.cgi?make(1)) says:

> The `$@` and `$<` variables are more or less universally portable, as is the `$(MAKE)` variable.

but then does not mention `$<` anywhere else on the page! Perhaps less universally portable...

shadowspawn:
Thanks @kaiquekandykoga

shadowspawn:
Released in v9.0.1

PR #748, opened 2022-12-09
Add information to README.md

KSGitUser:
# Pull Request

Add information in README.md

## Problem

After completing steps described in README.md, by creating folders and changing rights to them, user may stay on folder with no rights to create `n` file using `curl`. I think it would be better to change current directory before download data.

## Solution

Add command in README.md to change current folder to '/usr/local/n', which was created before

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ To avoid requiring `sudo` for `n` and `npm` global installs, it is suggested you
 
 If `npm` is not yet available, one way to bootstrap an install:
 
+    # In order to have rights to create 'n' file go to 'n' folder
+    cd /usr/local/n
     curl -L https://raw.githubusercontent.com/tj/n/master/bin/n -o n
     bash n lts
     # Now node and npm are available


shadowspawn:
I do not think these are the right instructions, but thanks for raising it. The user may not have followed the permissions steps, and in particular might be using `N_PREFIX` to install to their home folder and not have created `/usr/local/n`.

Downloading the script to _somewhere_, and then probably deleting it, are extra steps. I suggested downloading the file partly so there was an opportunity for the user to inspect the script before running it, but I see many other products just piping the script straight into a shell to keep the install as easy as possible.

```console
# https://brew.sh
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"

# https://github.com/mklement0/n-install
curl -L https://bit.ly/n-install | bash

# https://github.com/nvm-sh/nvm
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.2/install.sh | bash
```

I found various styles of combining curl and bash. The pipe is the most familiar and clean, but the `-c` approach allows sticking sudo in front of bash in the obvious way if needed.

```console
curl -fsSL https://raw.githubusercontent.com/tj/n/master/bin/n | bash -s lts
bash -c "$(curl -fsSL https://raw.githubusercontent.com/tj/n/master/bin/n)" -s lts
```

shadowspawn:
Reminder to self:
- the download of n is still useful as an "install" method for n itself
- the documentation for the install-node shortcut should say it is running n and not an install of n

shadowspawn:
Closing in favour of #754

Issue #673, opened 2021-05-23
Move implementation from Bash to POSIX shell?


shadowspawn:
# Problem

macOS no longer uses Bash as the default shell, and includes Bash v3 when the latest version is v5. Most environments will have a POSIX compatible shell, but not all environments have Bash available by default.

# Solution

`n` could be rewritten as a POSIX shell script?

rockyedge:
Rude but you know its just batch right i can give you 65 percent hit for a
favor

shadowspawn:
No feedback, closing. Something I still have in my backlog as a possible, but likely to be driven by need: an environment that people want to use that has sh and not bash.

Issue #648, opened 2021-01-19
Solution for n running on Alpine


immanuelfodor:
I think I've found the solution for running `n` on Alpine (see https://github.com/tj/n/issues/431).

I checked the official Docker image of Node at https://hub.docker.com/_/node

The latest LTS latest Alpine was that I followed (`14.15.4-alpine3.12, 14.15-alpine3.12, 14-alpine3.12, fermium-alpine3.12, lts-alpine3.12`), and it took me to this Dockerfile: https://github.com/nodejs/docker-node/blob/ad676318f09f9dee821ac6704340000d75fe31bc/14/alpine3.12/Dockerfile#L21

The highlighted line contains a link to `https://unofficial-builds.nodejs.org/download/release/v$NODE_VERSION/node-v$NODE_VERSION-linux-$ARCH-musl.tar.xz`

When I opened the link in my browser, I could download the XZ file: https://unofficial-builds.nodejs.org/download/release/v14.15.1/

Extracted it, and it contains the same structure than the non-Alpine NodeJS XZ distribution files: 

![image](https://user-images.githubusercontent.com/21174107/105067278-d375ef00-5a7f-11eb-84c0-9cbe25a9a70c.png)

When I check the build, it's indeed a `musl` build:

```bash
$ readelf -l node | grep "program interpreter"
      [Requesting program interpreter: /lib/ld-musl-x86_64.so.1]
```

In contrast, the version downloaded by `n` on Alpine now resolves to a non-existent dependency:

```bash
$ readelf -l /usr/local/bin/node | grep "program interpreter"
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
```

And it makes my CI pipeline die:

![image](https://user-images.githubusercontent.com/21174107/105067691-41221b00-5a80-11eb-92ee-92712787fab0.png)

The file is definitely there, what is missing it's runtime dependency as it was not built for Alpine: https://unix.stackexchange.com/questions/18061/why-does-sh-say-not-found-when-its-definitely-there

![image](https://user-images.githubusercontent.com/21174107/105067761-58610880-5a80-11eb-91c5-2b68a183d4cd.png)

Could you please adjust `n` to detect if it's running on Alpine, and then use the https://unofficial-builds.nodejs.org/ repo to download the `musl` build what works fine on Alpine? 

Although the repo is named unofficial, the official Docker image by Docker uses it, so we could do, too.

shadowspawn:
[...]

PR #751, opened 2022-12-14
Make script POSIX compliant (and switch shebang to sh)

shadowspawn:
# Pull Request

May not merge in absence of any user feedback, but wanted to check it was possible!

## Problem

`n` is currently written as a bash script. Bash isn't always available by default, and is an old version on macOS.

Related: #344 #648 #673 

## Solution

Rewrite as a POSIX shell script so runs with whatever (POSIX) shell is installed.

All tests pass on Ubuntu using `dash` as the shell. Have addressed all shellcheck issues with flags passed to utilities.

Currently still using `local` (a lot) which is commonly available but not POSIX.

https://stackoverflow.com/a/18600920/1082434

> It is normally done with the `local` keyword, which is, as you seem to know, not defined by POSIX. Here is an informative [discussion about adding 'local' to POSIX](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=381091).
>
> However, even the most primitive POSIX-compliant shell I know of which is used by some GNU/Linux distributions as the `/bin/sh` default, `dash` (Debian Almquist Shell), supports it. FreeBSD and NetBSD use `ash`, the original Almquist Shell, which also supports it. OpenBSD uses a `ksh` implementation for `/bin/sh` which also supports it. So unless you're aiming to support non-GNU non-BSD systems like Solaris, or those using standard `ksh`, etc., you could get away with using `local`.


shadowspawn:
Issue with "super barebones" setup and only sh available: #782

shadowspawn:
I haven't had enough reason to move from bash to pure posix to date, closing this draft.

Issue #306, opened 2015-09-17
n doesn't overwrite existing node version if trying to switch to the same version but just different architecture


cgadam:
Hi,

I'm running _n_ version _2.0.2_ and I'm getting this weird behavior: when trying to change node version to a version I already had installed but with different architecture _n_ does nothing. It just assumes that the version is already installed but, in fact, it's not because I need the same version but with a different architecture.

Seems like the architecture itself is not considered when checking if the version exists or not.

Example:

```
cgadam-mac01:public cgadam$ n --version
2.0.2
cgadam-mac01:public cgadam$ n 0.12.7

     install : node-v0.12.7
       mkdir : /Users/cgadam/n/n/versions/node/0.12.7
       fetch : https://nodejs.org/dist/v0.12.7/node-v0.12.7-darwin-x64.tar.gz
   installed : v0.12.7

cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.7
x64
cgadam-mac01:public cgadam$ n 0.12.4
cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.4
ia32
cgadam-mac01:public cgadam$ n -a x86 0.12.7
cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.7
x64 //Here I would expect ia32 instead!
```

There's a workaround to this which consists on deleting the previous version and installing from scratch:

```
cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.7
x64
cgadam-mac01:public cgadam$ n 0.12.4
cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.4
ia32
cgadam-mac01:public cgadam$ n rm 0.12.7
cgadam-mac01:public cgadam$ n -a x86 0.12.7

     install : node-v0.12.7
       mkdir : /Users/cgadam/n/n/versions/node/0.12.7
       fetch : https://nodejs.org/dist/v0.12.7/node-v0.12.7-darwin-x86.tar.gz
   installed : v0.12.7

cgadam-mac01:public cgadam$ node --version && node -p -e process.arch
v0.12.7
ia32
```

But it would be nice to consider the architecture part of the version! :+1:

maxrimue:
[...]

PR #753, opened 2022-12-20
Display node version architecture when on interactive mode

MatteoGauthier:
# Pull Request

Display node versions architecture when on interactive mode

## Problem

Sometimes npm libs can be compatible only with arm64 or x64

## Solution

```
n
```
Capture d’écran 2022-12-20 à 11 22 36

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -505,10 +505,11 @@ display_versions_with_selected() {
   local selected="$1"
   echo
   for version in $(display_versions_paths); do
+    current_version="$version - $(file $CACHE_DIR/$version/bin/node -b | awk '{print $4}')"
     if test "$version" = "$selected"; then
-      printf "  ${SGR_CYAN}ο${SGR_RESET} %s\n" "$version"
+      printf "  ${SGR_CYAN}ο${SGR_RESET} %s\n" "$current_version"
     else
-      printf "    ${SGR_FAINT}%s${SGR_RESET}\n" "$version"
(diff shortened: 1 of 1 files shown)


shadowspawn:
There is limited support in `n` for the architecture, and it has not come up often other than during the transition to having an Apple silicon native node available. One earlier issue is #306. 

An alternative to detecting the node binary type would be to store the architecture used for the download (in a file or in the folder naming or hierarchy).

A couple of comments specifically about code in this PR:

1) I would probably not want this displayed all the time, since for most situations it is not relevant.

2) The `file` command is not universally available, and the output may not be consistent. In particular, it is not installed in the Ubuntu container used in the tests.

```console
root@ea5b5e5fd71c:/# uname -a
Linux ea5b5e5fd71c 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
root@ea5b5e5fd71c:/# file
bash: file: command not found
```

MatteoGauthier:
@shadowspawn Thanks for your comment and code advices 🙏 I think it might be better to optionally show this information as you said

shadowspawn:
I haven't found a nice way of finding out architecture of node binary that works across multiple platforms.

We can revisit this if comes up again in issues, and have more information about problems people are encountering. Currently not a clear enough problem with a clean enough solution. 😅

Issue #762, opened 2023-03-20
Extra logging for when installed npm is not active after install, or extra test in doctor


googol7:
## Short Version
When I ran `n` it didn’t install the bundled `npm` version. I had to uninstall the globally installed `npm`. Now it works.

## Long Version
When I switched the Node version with `n` I just saw:

```
     copying : node/14.21.2
   installed : v14.21.2
```

I expected to see:
```
     copying : node/14.21.2
   installed : v14.21.2 (with npm 6.14.17)
```

I didn’t [preserve](https://github.com/tj/n#preserving-npm) the `npm` version.

`n doctor` didn’t tell me why.

# Solution
I uninstalled the globally installed `npm` to let `n` also switch to the bundled `npm` version:
`npm uninstall -g npm`

# Feature request
`n` should be a little bit more verbose and tell us why it doesn’t also switch the bundled `npm` version. The message could be:
```
     copying : node/14.21.2
   installed : v14.21.2
        hint : Didn’t switch the bundled `npm` version because it is installed globally.
```

The hint could also be:
        hint : Didn’t switch the bundled `npm` version because the environment variable `N_PRESERVE_NPM` is set.

## References
- https://github.com/tj/n/issues/721
- https://github.com/tj/n/issues/484
- https://github.com/tj/n/issues/374
- https://github.com/tj/n/issues/218
- https://github.com/tj/n/issues/620
- https://github.com/tj/n/issues/598
- https://github.com/tj/n/issues/587
- https://github.com/tj/n/issues/258
- https://github.com/tj/n/issues/329 

## Configuration Details

```bash
$ n --version
v9.0.1

$ command -v node
/usr/local/bin/node

$ node -p process.platform
darwin

$ n doctor
This information is to help you diagnose issues, and useful when reporting an issue.
Note: some output may contain passwords. Redact before sharing.


COMMAND LOCATIONS AND VERSIONS

bash
/bin/bash
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin22)
Copyright (C) 2007 Free Software Foundation, Inc.

n
/Users/phil/.npm/bin/n
v9.0.1

node
/usr/local/bin/node
v9.11.2
JavaScript engine: v8

npm
/usr/local/bin/npm
5.6.0

┌─────────────────────────────────────────────────────────┐
│                 npm update check failed                 │
│           Try running with sudo or get access           │
│          to the local update config store via           │
│ sudo chown -R $USER:$(id -gn $USER) /Users/phil/.config │
[...]

PR #764, opened 2023-04-07
Add npm check to doctor

shadowspawn:
## Problem

People may have installed npm separately before trying `n` for managing Node.js (node + npm) and shadow the version installed by `n`.

See: #762

## Solution

Add a check to `n doctor`. This could be detected at install time, but just adding to `n doctor` for now.

Generates warning like:
```
Checking npm install destination...
There is an active version of npm shadowing the version installed by n. Check what is in PATH (unless this is expected).
   installed : /usr/local/bin/npm
      active : /Users/john/my_npm/bin/npm
```

## ChangeLog

- check for possible problem with multiple `npm` locations when running `n doctor`

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1484,6 +1484,20 @@ function show_diagnostics() {
     fi
   fi
 
+  # Check npm too. Simpler check than for PATH and node, more like the runtime logging for active/installed node.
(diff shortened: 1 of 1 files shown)


shadowspawn:
Tracking new cases where an install time check might have helped:
- #765 (HomeBrew)

PR #544, opened 2018-12-20
add timeout for net request

zp1996:
sometimes in china will be timeout:)

PR #771, opened 2023-07-09
add curl --max-time

wangduanduan:
# Pull Request

## Problem

n seems to be stuck after  wait a very long time, finally i found out that curl timeout, i think it will be nice to set a short default timeout for curl

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ In brief:
 - `N_MAX_REMOTE_MATCHES` to change the default `ls-remote` maximum of 20 matching versions
 - `N_PRESERVE_NPM`: See [Preserving npm](#preserving-npm)
 - `N_PRESERVE_COREPACK`: See [Preserving npm](#preserving-npm)
+- `N_CURL_MAX_TIMEOUT`: See curl **--max-time**
 
 ## How It Works
 
--- a/bin/n
+++ b/bin/n
@@ -109,7 +109,7 @@ GET_SHOWS_PROGRESS="false"
 # --location to follow redirects
 # --fail to avoid happily downloading error page from web server for 404 et al
 # --show-error to show why failed (on stderr)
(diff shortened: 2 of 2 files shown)


shadowspawn:
Related: #544

shadowspawn:
There have not been many issues opened about timeouts. I am reluctant to set a custom default when might cause timeouts that given a bit longer might have worked. To some extent we would be picking a value that works for one person.

Would the 5s in the PR be a suitable default for all your curl calls? In which case it could go in default curl config file: https://everything.curl.dev/cmdline/configfile

wangduanduan:
In china , [nodejs.org](http://nodejs.org/) is often inaccessible, which causes us to waste a lot of time trying to troubleshoot the real reason.

Setting up curlrc assumes that the issue is with the network, but it does not necessarily reduce the time spent troubleshooting the problem.

shadowspawn:
Ah right, problem is the node mirror not all use of `curl`.

Another relevant setting which might improve behaviour is using a custom mirror. I haven't heard about how well it is working recently, but it is the example in the README:  https://github.com/tj/n#custom-source

```
export N_NODE_MIRROR=https://npmmirror.com/mirrors/node
```

shadowspawn:
> I am reluctant to set a custom default when might cause timeouts that given a bit longer might have worked. To some extent we would be picking a value that works for one person.

`5 s` is too short. I sometimes see downloads taking 20-30 s. Not sure if the server is sometimes slow, or I get throttled when running lots of downloads for `n` unit tests.

Let us know how the mirror site works.

shadowspawn:
Does using `npmmirror.com` improve your timeout problems @wangduanduan ?

wangduanduan:
yes, it is

wangduanduan:
> Does using `npmmirror.com` improve your timeout problems @wangduanduan ?

Yes, it is

shadowspawn:
Thanks for confirming @wangduanduan 

I think using the country local mirror is a better high-level answer than changing the timeout in this case.

Issue #726, opened 2022-05-21
Using original node version if no file found for auto version


artfulsage:
Hey! I like the idea of `auto` version, it's really cool and gives us some great posibilities. But I think that it would be nice to let use it not only inside nodejs-project directories but everywhere. So I suggest to use original (current installed) nodejs version if no file was found for auto version. For me it fits in `auto` concept. What do you think? I would like to provide a PR if you agree.

shadowspawn:
I don't understand the scenario you are describing. What command are you running with `auto` where silently continuing to use current version would be desirable?

`n` currently displays an error for `auto` and `engine` if no control file found so the user knows that no version was implied, and might need to think about what folder they are in or what version of node they want to use.

artfulsage:
I understand the current behaviour the same way as you described it. But this new behavior would make possible to create an alias `node = n run auto` and so we could use the node version dependent on current folder transparantly. Actually we will get a multiple installed versions behavior. Yes, there are limitations with npm version installed, but for experienced users and `yarn`-users I guess it wouldn't be a problem.

Also I see possible future improvement when there wouldn't be any need to copy current node binaries to `/usr/local/bin/node`. `n` should just store version of current pseudo-installed version in some file and use it inside `auto`-resolver. Still there is a problem with npm version but perhaps it could be solved some way with some aliases.

shadowspawn:
Lots of comments!

I don't want to change the current behaviour which expects a configured auto version, based on current feedback. There have not been (other) requests, and in fact one issue raised because `n` was not showing an error: https://github.com/tj/n/issues/692#issuecomment-931848225

There could be an option to opt-in to this behaviour, or a different label. At the moment I don't think I would want to add it without this issue getting upvoted.

The fallback could run the installed version, or could run the matching version from the cache. The results are slightly different. I think the installed version might match the semantics of your intended usage better, but the cached version matches the semantics of the `run` command better.

This could be externally scripted by using `n ls-remote auto` to look for a configured version. That would let you try out the usage without needing to modify `n`.

> Also I see possible future improvement when there wouldn't be any need to copy current node binaries

`n run` and `n exec` do not copy the binaries.

> Still there is a problem with npm version

`npm exec` gives access to the `npm` version included in the target Node.js version. So in line with your alias for node, could have `auto-npm = n exec auto npm`.

artfulsage:
[...]

PR #806, opened 2024-07-20
Add optional fallback for `auto` and `engine`

elibarzilay:
This is a sketch commit that makes it possible to use, eg, `auto:lts` to mean "same as `auto`, or `lts` if that didn't work". With this you can get the benefit of `auto` when applicable, or `lts` otherwise.

An alternative approach would be to use an envvar, as in

    N_FALLBACK_VERSION=lts n auto

but that would be less convenient.

---

An example of a cute hack is to create an executable file that contains

    #!/bin/bash
    exec n -q -d exec auto:lts "$(basename "$0")" "$@"

then symlink it as `/usr/local/bin/node` etc, and now you have a magical `node` command that follows whatever version is specified in your current directory. (This fails in case of running with `engine`, since then requiring node itself means an infinite loop, but could be done with some more tweaks.)

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1076,7 +1076,7 @@ function get_file_node_version() {
 }
 
 #
-# Synopsis: get_package_engine_version\
+# Synopsis: get_package_engine_version
 # Sets g_target_node
 #
 
@@ -1142,13 +1142,14 @@ function get_nvmrc_version() {
 }
 
 #
-# Synopsis: get_engine_version [error-message]
+# Synopsis: get_engine_version [default-version]
 # Sets g_target_node
+# Uses error_message, if set
 #
 
 function get_engine_version() {
   g_target_node=
-  local error_message="${1-package.json not found}"
+  local error_message="${error_message-package.json not found}"
   local parent
   parent="${PWD}"
   while [[ -n "${parent}" ]]; do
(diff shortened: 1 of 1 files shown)


shadowspawn:
Interesting idea. However, I don't think this is a common enough use case to justify the complication. 

You can get simple fallback behaviour by having a `.node-version` file in your home directory.

There are heavier-weight utilities that focus on per-directory tooling, like `direnv`, `asdf`, `mise`, and `volta`.

Somewhat related: #726

elibarzilay:
@shadowspawn consider that:

1. The actual change is pretty simple, just the additional argument to the two functions, and the dirt-simple "parsing".
2. Sounds like you're considering that cute hack as my use case, but it's more than that.  Consider cases where you cannot have a reliable `.node-version` because you're doing a build in some random place (common in CI scripts, running on random configurations).
3. More than that, assume that you're writing some script that gets to run across multiple projects, some with a version specified, and some without it.  Going through the hack in #726 is solving that in an indirect and hacky way, not to mention going through `n` an extra time.

This is the reason I wanted something like this, enough that I might just fork it for my use case...

(BTW, these other tools are not fitting the nice place of `n` -- specifically not anything that is more generalized than just a node version, or things that rely on shell functions instead of an actual script. (I'm discouraging uses of `nvm` in CI pipelines because of that...))

shadowspawn:
If you would like to open an issue describing the suggestion, I am happy to leave issues open for months to gather likes and feedback.

PR #810, opened 2024-08-03
If `jq` is installed, use it to get values from package.json

### LINKED ISSUE #807: Finding `engine` is pretty slow
Since it uses `node` to find the version, and starting it is kind of slow.

If it's ok to use `jq`, I'll be happy to make a PR that uses it if it exists, or use `node` otherwise.

elibarzilay:
# Pull Request

`jq` is a popular choice for being installed, especially on CI machines, and using it is much faster than running node.

## Problem

Using `node` can be problematic if it's not installed yet, and it also has a much higher overhead (= time) to start.  Doesn't matter for an automated CI case, but in interactive use the potential doubling of running `node` is noticeable.

## Solution

Use `jq` if it's installed.  This is very straightforward use of `jq`, so prefer that over `node` if it is installed.

## ChangeLog

* Use `jq` to read values from `package.json` if it is installed.

Closes #807.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1084,9 +1084,14 @@ function get_package_engine_version() {
   g_target_node=
   local filepath="$1"
   verbose_log "found" "${filepath}"
-  command -v node &> /dev/null || abort "an active version of node is required to read 'engines' from package.json"
   local range
-  range="$(node -e "package = require('${filepath}'); if (package && package.engines && package.engines.node) console.log(package.engines.node)")"
+  if command -v jq &> /dev/null; then
+    range="$(jq -r '.engines.node // ""' < "${filepath}")"
+  elif command -v node &> /dev/null; then
+    range="$(node -e "package = require('${filepath}'); if (package && package.engines && package.engines.node) console.log(package.engines.node)")"
+  else
+    abort "either jq or an active version of node is required to read 'engines' from package.json"
+  fi
   verbose_log "read" "${range}"
   [[ -n "${range}" ]] || return 2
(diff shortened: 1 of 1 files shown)


bin/n
```diff
@@ -1474,6 +1479,13 @@ function show_diagnostics() {
     echo_red "Neither curl nor wget found. Need one of them for downloads."
   fi
 
+  printf "\njq\n"
+  if command -v jq &> /dev/null; then
+    command -v jq && jq --version
+  else
+    echo "jq not found, can be used to get package.json values faster than node"
```
shadowspawn, line 1486:
```suggestion
    echo "jq not found. Optional install used for reading engine version from package.json."
```

shadowspawn (approved):
I think the combination of a speed boost and allowing use of `n engine` without node installed make this worthwhile.

LGTM. One wording suggestion.

bin/n
elibarzilay: Yeah, I don't mind the phrasing -- I forgot to say that you should make it whatever fits...

MikeMcC399:
@shadowspawn 

Thanks for this great package!

Please label [10.0.0](https://github.com/tj/n/releases/tag/v10.0.0) as `latest` release so it shows up as such on https://github.com/tj/n/releases. Currently [9.2.0](https://github.com/tj/n/releases/tag/v9.2.0) holds the `latest` label.

shadowspawn:
Labelled as latest now on GitHub releases page. Thanks.

PR #815, opened 2024-09-27
mask download url when there is a password

jinxiao:
# Pull Request

mask download url when there is a password.

## Problem

When I use export N_NODE_MIRROR with username and password, I found that the password is a plaintext which cause our security concern.
## Solution

I introduce a new variable called masked_url, when called log or something want to output the url, can use this variable.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -793,7 +793,8 @@ install() {
   log installing "${g_mirror_folder_name}-v$version"
 
   local url="$(tarball_url "$version")"
-  is_ok "${url}" || abort "download preflight failed for '$version' (${url})"
+  local masked_url=$(echo "$url" | sed -r 's/(https?:\/\/[^:]+):([^@]+)@/\1:*****@/')
+  is_ok "${url}" || abort "download preflight failed for '$version' (${masked_url})"
 
   log mkdir "$dir"
   mkdir -p "$dir" || abort "sudo required (or change ownership, or define N_PREFIX)"
@@ -801,7 +802,7 @@ install() {
 
   cd "${dir}" || abort "Failed to cd to ${dir}"
 
-  log fetch "$url"
+  log fetch "${masked_url}"
   do_get "${url}" | tar "$tarflag" --strip-components=1 --no-same-owner -f -
   pipe_results=( "${PIPESTATUS[@]}" )
   if [[ "${pipe_results[0]}" -ne 0 ]]; then
--- a/bin/n
+++ b/bin/n
@@ -1498,7 +1498,7 @@ function show_diagnostics() {
   printf "\n\nSETTINGS\n"
 
   printf "\nn\n"
-  echo "node mirror: $(display_masked_url ${N_NODE_MIRROR})"
+  echo "node mirror: $(display_masked_url "${N_NODE_MIRROR}")"
   echo "node downloads mirror: $(display_masked_url "${N_NODE_DOWNLOAD_MIRROR}")"
   echo "install destination: ${N_PREFIX}"
   [[ -n "${N_PREFIX}" ]] && echo "PATH: ${PATH}"
--- a/bin/n
+++ b/bin/n
@@ -1245,7 +1245,7 @@ display_remote_index() {
   do_get_index "${index_url}" | tail -n +2 | cut -f 1,3,10
   if [[ "${PIPESTATUS[0]}" -ne 0 ]]; then
     # Reminder: abort will only exit subshell, but consistent error display
-    abort "failed to download version index $(display_masked_url "${index_url}")"
+    abort "failed to download version index ($(display_masked_url "${index_url}"))"
   fi
 }
 
--- a/bin/n
+++ b/bin/n
@@ -797,7 +797,7 @@ install() {
   log installing "${g_mirror_folder_name}-v$version"
 
   local url="$(tarball_url "$version")"
-  is_ok "${url}" || abort "download preflight failed for '$version' '$(display_masked_url ${url})'"
+  is_ok "${url}" || abort "download preflight failed for '$version' '$(display_masked_url "${url})"'"
 
   log mkdir "$dir"
   mkdir -p "$dir" || abort "sudo required (or change ownership, or define N_PREFIX)"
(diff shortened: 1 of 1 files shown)


shadowspawn:
I am open to this. There are quite a log of places an URL gets displayed. One place I am particularly aware of is in `show_diagnostics`, enough that I put in a warning `Note: some output may contain passwords. Redact before sharing.`.

I suggest move the masking code into a function, say `display_masked_url`, so it can be used in more places.

jinxiao:
> I am open to this. There are quite a log of places an URL gets displayed. One place I am particularly aware of is in `show_diagnostics`, enough that I put in a warning `Note: some output may contain passwords. Redact before sharing.`.
> 
> I suggest move the masking code into a function, say `display_masked_url`, so it can be used in more places.

Good idea, I have changed this. Could you please have a check?  I also changed the download URL in diagnostics to avoid any password leakage.

bin/n
```diff
@@ -793,15 +797,15 @@ install() {
   log installing "${g_mirror_folder_name}-v$version"
 
   local url="$(tarball_url "$version")"
-  is_ok "${url}" || abort "download preflight failed for '$version' (${url})"
+  is_ok "${url}" || abort "download preflight failed for '$version' '$(display_masked_url ${url})'"
```
shadowspawn, line 800: The password may have a space in it, so the url needs to be quoted to pass the expansion as a single argument:
```suggestion
  is_ok "${url}" || abort "download preflight failed for '$version' '$(display_masked_url "${url})"'"
```

(Applies to other uses too.)

bin/n
```diff
@@ -1494,8 +1498,8 @@ function show_diagnostics() {
   printf "\n\nSETTINGS\n"
 
   printf "\nn\n"
-  echo "node mirror: ${N_NODE_MIRROR}"
-  echo "node downloads mirror: ${N_NODE_DOWNLOAD_MIRROR}"
+  echo "node mirror: $(display_masked_url ${N_NODE_MIRROR})"
+  echo "node downloads mirror: $(display_masked_url "${N_NODE_DOWNLOAD_MIRROR}")"
```
shadowspawn, line 1502: (This one has quotes already.) 👍

bin/n
```diff
@@ -793,15 +797,15 @@ install() {
   log installing "${g_mirror_folder_name}-v$version"
 
   local url="$(tarball_url "$version")"
-  is_ok "${url}" || abort "download preflight failed for '$version' (${url})"
+  is_ok "${url}" || abort "download preflight failed for '$version' '$(display_masked_url "${url})"'"
```
shadowspawn, line 800:
```suggestion
  is_ok "${url}" || abort "download preflight failed for $version $(display_masked_url "${url}")"
```

bin/n
```diff
@@ -793,15 +797,15 @@ install() {
...
 
   log mkdir "$dir"
   mkdir -p "$dir" || abort "sudo required (or change ownership, or define N_PREFIX)"
   touch "$dir/n.lock"
 
   cd "${dir}" || abort "Failed to cd to ${dir}"
 
-  log fetch "$url"
+  log fetch "$(display_masked_url ${url})"
```
shadowspawn, line 808:
```suggestion
  log fetch "$(display_masked_url "${url}")"
```

bin/n
```diff
@@ -1241,7 +1245,7 @@ display_remote_index() {
   do_get_index "${index_url}" | tail -n +2 | cut -f 1,3,10
   if [[ "${PIPESTATUS[0]}" -ne 0 ]]; then
     # Reminder: abort will only exit subshell, but consistent error display
-    abort "failed to download version index (${index_url})"
+    abort "failed to download version index $(display_masked_url ${index_url})"
```
shadowspawn, line 1248:
```suggestion
    abort "failed to download version index $(display_masked_url "${index_url}")"
```

bin/n
```diff
@@ -1494,8 +1498,8 @@ function show_diagnostics() {
   printf "\n\nSETTINGS\n"
 
   printf "\nn\n"
-  echo "node mirror: ${N_NODE_MIRROR}"
-  echo "node downloads mirror: ${N_NODE_DOWNLOAD_MIRROR}"
+  echo "node mirror: $(display_masked_url ${N_NODE_MIRROR})"
```
shadowspawn, line 1501:
```suggestion
  echo "node mirror: $(display_masked_url "${N_NODE_MIRROR}")"
```

shadowspawn:
The quotes are a bit tricky with the nesting!

shadowspawn:
(Closed by mistake.)

bin/n
```diff
@@ -793,15 +797,15 @@ install() {
   log installing "${g_mirror_folder_name}-v$version"
 
   local url="$(tarball_url "$version")"
-  is_ok "${url}" || abort "download preflight failed for '$version' (${url})"
+  is_ok "${url}" || abort "download preflight failed for $version $(display_masked_url "${url}")"
```
shadowspawn, line 800: Preserving original style now I realise the single quotes and brackets appear in the output.
```suggestion
  is_ok "${url}" || abort "download preflight failed for '$version' ($(display_masked_url "${url}"))"
```

bin/n
```diff
@@ -1241,7 +1245,7 @@ display_remote_index() {
   do_get_index "${index_url}" | tail -n +2 | cut -f 1,3,10
   if [[ "${PIPESTATUS[0]}" -ne 0 ]]; then
     # Reminder: abort will only exit subshell, but consistent error display
-    abort "failed to download version index (${index_url})"
+    abort "failed to download version index $(display_masked_url "${index_url}")"
```
shadowspawn, line 1248:
```suggestion
    abort "failed to download version index ($(display_masked_url "${index_url}"))"
```

shadowspawn:
I misunderstood a couple of the uses, so added suggestions to preserve the original style.

shadowspawn:
I was worried about spaces in password, but now I remember username and password need to be url-encoded for robustness anyway: https://github.com/tj/n/blob/master/docs/proxy-server.md

However, adding the quotes around the variable expansions fixes warnings from `shellcheck` so I would have wanted the changes anyway.

shadowspawn:
Thanks @jinxiao

shadowspawn:
Released in v10.1.0

PR #826, opened 2024-12-19
Add option to open a subshell with desired nodejs version

gergof:
# Pull Request

## Problem

I frequently have to work with multiple nodejs versions at the same time (for example frontend works with nodejs 20, backend need 16). `n exec 16 bash` works, but then I don't know what nodejs version is used in that terminal window.

## Solution

Added a new command `n shell <version>` which modifies the PATH and PS1

## ChangeLog

Added new command to open subshell with specific nodejs version

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ wish to use node and npm, or are switching to a different way of managing them.
 
 ## Using Downloaded Node.js Versions Without Reinstalling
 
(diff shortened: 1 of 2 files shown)


shadowspawn:
This is interesting, but the shell and prompt are personal preference. I don't think this will suit enough users.

Issue #830, opened 2025-02-06
Feature request: architecture env var


elibarzilay:
Something like `$N_NODE_MIRROR` is useful to set in an environment that requires it as a default.  It would be nice to be able to do the same with some `$N_ARCH`.

Not having this is not only ugly (I use an env var for one setting, and a flag for the other), but less convenient since I cannot set a default as I can with the mirror.

In my case, I have an old server that uses the unofficial-builds repo to get the `x64-glibc-217` builds. The thing is that `n` has very simple logic for getting the architecture, so adding stuff that inspects the glibc version would probably be prohibitively complicate and even more impractical since the architecture can depend on the mirror that is used.

(I'm willing to do a PR if this sounds like a good idea.)

shadowspawn:
Thanks for the use case. 

Supporting `N_ARCH` would be in line with the mirror support and reasonable. I am open to looking at a PR for that (and willing to do it myself too). I think the code change might be a one liner, but I would also like `N_ARCH` listed in `n doctor` if set, like `N_PREFIX` et al.

elibarzilay:
Done in #832.  As I said there, the big chunk is just avoiding the `uname` dance if it's not needed, and it would be better to do that in a separate PR if you prefer that.

shadowspawn:
Released in 10.2.0. Thanks @elibarzilay

PR #832, opened 2025-02-08
Add `$N_ARCH` as a default architecture

elibarzilay:
# Pull Request

More convenient to use than --arch in case of a mirror with exotic architecture, like the ones in `unofficial-builds`.

Do this by making it the default for `$ARCH`.

Also, skip the `uname -m` if it's not needed because of this, or because of `--arch`, or the version-specific default. (This is not needed for the functionality.)

## Problem

See #830.

## Solution

The actual fix is just:

```diff
-ARCH=
+ARCH="$N_ARCH"
```

The rest are:
* README changes
* Ensuring that `$N_ARCH` exists if unset
* Skipping the `uname -m` if `$ARCH` is set (this could be done in a separate PR)
* Reporting `$N_ARCH` in `show_diagnostics`

## ChangeLog

New `$N_ARCH` to set a default architecture.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,11 +266,13 @@ Another example is the Node.js [unofficial-builds project](https://github.com/no
 
     export N_NODE_MIRROR=https://unofficial-builds.nodejs.org/download/release
 
-You may need to specify the architecture explicitly if not autodetected by `n`, such as using `musl` `libc` on Alpine:
+You may need to specify the architecture explicitly if not autodetected by `n`, such as using `musl` `libc` on Alpine. You can do that with `N_ARCH` or `--arch`:
 
     export N_NODE_MIRROR=https://unofficial-builds.nodejs.org/download/release
+    export N_ARCH=x64-musl
     apk add bash curl libstdc++
-    n --arch x64-musl install lts
+    n install lts
+    n --arch x64-musl install lts   # using a flag
 
 If the custom mirror requires authentication you can add the [url-encoded](https://urlencode.org) username and password into the URL. e.g.
 
--- a/bin/n
+++ b/bin/n
@@ -134,7 +134,7 @@ g_target_node=
 
 DOWNLOAD=false # set to opt-out of activate (install), and opt-in to download (run, exec)
 CLEANUP=false # remove cached download after install
-ARCH=
+ARCH="${N_ARCH-}"
 SHOW_VERBOSE_LOG="true"
 OFFLINE=false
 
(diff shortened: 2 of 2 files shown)


bin/n
```diff
@@ -1030,20 +1033,26 @@ function display_tarball_platform() {
...
-      # e.g. armv6l, armv7l, arm64
-      arch="${uname_m}"
-      ;;
-  esac
-    # Override from command line, or version specific adjustment.
-  [ -n "$ARCH" ] && arch="$ARCH"
+  # Set to $ARCH which is from
+  # * the command line,
+  # * otherwise, from $N_ARCH
+  # * otherwise from version specific adjustment (if applicable, see update_arch_settings_for_version)
```
shadowspawn, line 1039:
```suggestion
  # architecture might already be known from (priority order):
  # * --arch on command line
  # * N_ARCH
  # * version specific adjustment (see update_arch_settings_for_version)
```

bin/n
```diff
@@ -1030,20 +1033,26 @@ function display_tarball_platform() {
...
+  local arch="$ARCH"
+  if [[ -z "$arch" ]]; then
+    # if none of the above, infer from uname
```
shadowspawn, line 1042: I think just leave this comment out.

README.md
```diff
@@ -266,10 +266,15 @@ Another example is the Node.js [unofficial-builds project](https://github.com/no
 
     export N_NODE_MIRROR=https://unofficial-builds.nodejs.org/download/release
 
-You may need to specify the architecture explicitly if not autodetected by `n`, such as using `musl` `libc` on Alpine:
+You may need to specify the architecture explicitly if not autodetected by `n`, such as using `musl` `libc` on Alpine. You can do that with using `N_ARCH`:
```
shadowspawn, line 269: I think fine to just have `N_ARCH` covered in example here:

```suggestion
You may need to specify the architecture explicitly if not autodetected by `n`, such as using `musl` `libc` on Alpine. You can do that with `N_ARCH` or `--arch`:
```

README.md
```diff
@@ -266,10 +266,15 @@ Another example is the Node.js [unofficial-builds project](https://github.com/no
...
 
     export N_NODE_MIRROR=https://unofficial-builds.nodejs.org/download/release
+    export N_ARCH=x64-musl
     apk add bash curl libstdc++
+    n install lts
+
+or using `--arch`:
+
     n --arch x64-musl install lts
```
shadowspawn, line 278: Can leave out the `--arch` usage here.

bin/n
```diff
@@ -76,6 +76,9 @@ N_NODE_MIRROR=${N_NODE_MIRROR:-${NODE_MIRROR:-https://nodejs.org/dist}}
 N_NODE_MIRROR=${N_NODE_MIRROR%/}
 readonly N_NODE_MIRROR
 
+N_ARCH="${N_ARCH-}"
+readonly N_ARCH
+
```
shadowspawn, line 81: Perhaps leave this out? In the other cases like `N_NODE_MIRROR`, the global variable is set to a default value if not supplied and/or normalised. Nothing extra to do with `N_ARCH`.

There are other globals (from env) that are just used in the code like `N_PRESERVE_COREPACK`.

(I have wondered about having all the globals mentioned here, but have not established that as a pattern.)

shadowspawn:
Thanks. Lots of cosmetic comments. I'll do some testing before approving.

bin/n
elibarzilay: OK.  TBH, I couldn't figure out if you'd prefer a consistent declaration thing or not.  (And it actually had an effect on the code: I wanted to set it due to the version thing, but didn't do that to keep it readonly.)

In any case, I removed it.

README.md
elibarzilay: (Left a comment at the end of the line, to avoid the illusion that both are needed...)

README.md
elibarzilay: Combined both, see the new patch which I'll push shortly.

elibarzilay:
Made all changes.  (You might have more comments though, so not to be taken blindly...)

Jose1654 (approved):

bin/n
HyphensPeciwse33: #850

HyphensPeciwse33:
Duplicate of #10

PR #838, opened 2025-07-03
Ensure npm and corepack are copied if missing

Hopsken:
Enhances logic to ensure npm and corepack are cleanly copied when missing, even if preservation variables are set, improving consistency of toolchain setup.

# Pull Request

## Problem

If `N_PRESERVE_NPM` and `N_PRESERVE_COREPACK` are set before any node versions installed, it will never be copied.

## Solution

Ignore `N_PRESERVE_COREPACK` and `N_PRESERVE_NPM` if it's never get set before and copy anyway.

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -698,13 +698,13 @@ activate() {
   mkdir -p "$N_PREFIX/lib"
   # Copy everything except node_modules.
   find "$dir/lib" -mindepth 1 -maxdepth 1 \! -name node_modules -exec cp -fR "{}" "$N_PREFIX/lib" \;
-  if [[ -z "${N_PRESERVE_NPM}" ]]; then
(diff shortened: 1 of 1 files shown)


shadowspawn:
I know I used the word "preserve" in the naming and description, but I am comfortable that the implementation is the same as "skip". I originally chose "preserve" to emphasise it was useful to npm users who were manually installing the latest version of npm and running ahead of the embedded version, which was the use case I first hit.

The other significant use case is people who are using a different package manager, and have no need for npm at all. The "preserve" naming is not great there but the functionality is fine.

> If `N_PRESERVE_NPM` and `N_PRESERVE_COREPACK` are set before any node versions installed, it will never be copied.

I am not seeing this as a significant problem. Is your use case the slight complication of installing npm before using npm to install a later npm? The existing work-around is to override the environment variables for the first install:

```
n --no-preserve lts
```

Hopsken:
Thanks for the clear explanation.

Issue #825, opened 2024-12-15
[feature request] add the previous version in the output.


vdegenne:
I like to run `sudo n latest` every now and then to make sure I'm always on the last version of NodeJS. The problem with `n` is that it's not always apparent whether node got updated or not. And if it does it doesn't necessarily show what upgrade it got.
In the output it would be nice to also show the previous version, e.g.

```
  installing : node-v23.4.0
       mkdir : /usr/local/n/versions/node/23.4.0
       fetch : https://nodejs.org/dist/v23.4.0/node-v23.4.0-linux-x64.tar.xz
     copying : node/23.4.0
   installed : v23.4.0 (with npm 10.9.2)
   previous  : v23.3.0
```

shadowspawn:
> The problem with n is that it's not always apparent whether node got updated or not.

I am not sure what this means. But I'll clarify what you see when you install.

When the target version is not already in the cache, like when a new version comes out, it is downloaded into the cache and installed (copied) into `/usr/local` (or `N_PREFIX`).

```console
$ n latest
  installing : node-v23.4.0
       mkdir : /usr/local/n/versions/node/23.4.0
       fetch : https://nodejs.org/dist/v23.4.0/node-v23.4.0-darwin-arm64.tar.xz
     copying : node/23.4.0
   installed : v23.4.0 (with npm 10.9.2)
```

When the target version is already in the cache, it is copied into `/usr/local` (or `N_PREFIX`). No download required.

```console
$ n ls
node/18.20.4
node/20.18.0
node/22.12.0
node/23.4.0

$ n latest
     copying : node/23.4.0
   installed : v23.4.0 (with npm 10.9.2)
```

We can also see/do the download and install separately for a new version:
```
$ n rm latest      
latest (23.4.0) not in downloads cache
$ n download latest
    download : node-v23.4.0
       mkdir : /usr/local/n/versions/node/23.4.0
       fetch : https://nodejs.org/dist/v23.4.0/node-v23.4.0-darwin-arm64.tar.xz
$ n install latest
     copying : node/23.4.0
   installed : v23.4.0 (with npm 10.9.2)
```

vdegenne:
Thanks for the details. 
I just meant to add the previous version it got updated from, for instance if I am on node v22.0.0 and use `n latest` it should show `v22.0.0` somewhere before telling it got updated to `v23.0.0` or something.

That's really just like adding `node --version` before anything got updated.

vdegenne:
[...]

Issue #727, opened 2022-05-26
Allow version matching against local downloaded versions, instead of always matching remote versions


hubertnnn:
# Bug Report

_Edit: original issue title was "n should use latest available version of node if installation is impossible"._

## Summary

When a new version of node gets released then n will start throwing errors instead of using latest available version

## Steps to Reproduce

```
# Assuming 14.19.3 is the latest version of branch 14
# Uninstall latest version
sudo n rm 14.19.3
# Install a slightly older one
sudo n 14.19.2
# Run an application using latest version of v14 branch
n exec 14 node --version
```

## Expected Behaviour

```
v14.19.2
```

Application should run showing `v14.19.2`, since that is the latest installed version of node 
and you don't have permissions to install a newer one (not root).

## Actual Behaviour

```
Error: '14' (14.19.3) not in downloads cache
```

## Other Information

I am not sure if choosing between downloading and using preinstalled node should be done automatically.
Possibly `n exec` should never download node since in most cases it will not be possible either way 
(`n exec <app>` should never be run as root and only root can install new node version)

Maybe having a switch to `download/use local` would be an option here as well.

## Configuration Details

```bash
$ n --version
8.2.0

$ command -v node
/usr/local/bin/node

$ node -p process.platform
linux
```

shadowspawn:
Short version: working as intended

Long version

The current behaviour of version lookups is to consult the available downloads, rather than the local cached versions.

By default `n run` and `n exec` only use cached versions and do not perform a download. The message you saw about "not in downloads cache " was not related to permissions.

There isn't currently an option to use the latest matching cached version. Another scenario that was raised in an earlier issue is no internet access: #404.

You can check what versions are available locally with `n ls`.

ikrong:
When there is version 14 on the user's computer, even if it is not the latest version, the local version should be used first when executing `n exec 14`, and an error should not be reported. A warning can be prompted; when there is no version 14 on the user's computer, an error should be reported when executing n exec 14.

local version 14.19.1
remote latest version 14.19.3

```shell
n exec 14 npm -v
```

expect behaviour

```shell
6.14.12
```

actual behaviour

```shell
Error: '14' (14.19.3) not in downloads cache
```

Both `n run` and `n exec` should prefer the local version

shadowspawn:
I don't think it likely I will change the default behaviour. There could be a new option to change the matching from "latest version according to Node.js" to "latest version already downloaded", as suggested in original post:

[...]

Issue #574, opened 2019-07-20
"n lts" downgrading npm version 6.10.1 to 6.9.0


pdh0710:
# Bug Report

## Summary

(Please excuse my English)

- I'm using node.js LTS version ( by executing `n lts` )
Currently node.js LTS version is v10.16.0.
- I manually upgrade npm to version 6.10.1 ( by executing `npm -g install npm` )

I just upgraded 'n' to version 5.0.1, execute `n lts`, and I found that 'npm' version was downgraded from 6.10.1 to 6.9.0.

At first, I did not know why. But, through the iterative test, I could found that executing `n lts` cause the problem. Finally I found that **node.js@10.16.0 is repeatedly installed** when I execute `n lts`.

Additionally, I found that the installation paths of node.js and global modules were changed from ['/usr/bin/node', '/usr/lib/node_modules'] to ['/usr/local/bin/node', '/usr/local/lib/node_modules']. I do not know the reason. But, I re-installed node.js global modules for safety. So two node.js('/usr/bin/node' and '/usr/local/bin/node', all v10.16.0) are installed in my system, currently.

Previous version of 'n' did not make these problems, so I think this is a newly appeared bug.

## My system informations

```bash
$ n --version
5.0.1

$ node -v
v10.16.0

$ which node
/usr/local/bin/node

$ npm root -g
/usr/local/lib/node_modules

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic

$ uname -a
Linux server 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

shadowspawn:
What has changed in `n` v5.0.0 is that when you run `n its` it now *always* does an install so you get consistent behaviour. Previously, if the active version of node was already the lts version then `n` would silently skip the install.

The install includes both node and npm. This can mean a downgrade of npm if you separately updated npm separately in the default install location. (You could change the npm prefix if you want to manage it yourself.)

The default install location used by `n` is to `/usr/local`. The other location you see will be installed by the Ubuntu package manager. It is confusing having two versions installed, so if you are going to use `n` and install to `/usr/local` then you might want to delete the "other" ones.

pdh0710:
[...]

Issue #543, opened 2018-12-16
Feature suggestion: have "install" say what it does


shadowspawn:
(Opening an issue for opportunity for feedback.)

## Problem

Running an install does not always do anything, and there is no feedback to inform the user (<https://github.com/tj/n/issues/198>). The install is silently skipped if the active version of node matches the requested version, and an activation is likewise silent so not visibly different.

This might be reasonable if only node was involved, but there is npm too. I personally find this problematic when intending to restore to a shipping combination of node+npm, or trying to recover or give recovery instructions from a broken npm.

### Uncertain outcome

````bash
n lts
# Did anything happen? What version of npm is active now?
````

- If you had a different version of node active, you now have the npm that ships with node lts
- If node lts was already active, then nothing has changed and npm has not changed from whatever you had before (which you may have manually updated, or broken)

### Extra steps to get predictable outcome

If you have a broken npm, or want someone to get to a known state, it takes two instructions. For example, change to a dummy version of node so that the second instruction actually does what is intended. (When I found myself doing this I decided it indicated a problem!)

````bash
n 8.0.0 # Any other version!
n 10.14.2
````

## Proposed Changes

- always install (i.e. remove optimization <https://github.com/tj/n/pull/128> skipping reinstall)
- include same output for activation as from full install

````bash
$ n 10.14.2

     install : node-v10.14.2
       mkdir : /usr/local/n/versions/node/10.14.2
       fetch : https://nodejs.org/dist/v10.14.2/node-v10.14.2-darwin-x64.tar.gz
   installed : v10.14.2

$ n 10.14.2
   installed : v10.14.2
````

As a bonus and following the same pattern, include extra output if active version of node is different than installed (due to PATH problems). This should help with a common problem case that trips people up and causes help requests (#213 #261 #290 #300 #367 #381 #400 #451).

````bash
$ N_PREFIX=~/deleteme n 10.14.1

     install : node-v10.14.1
       mkdir : /Users/john/deleteme/n/versions/node/10.14.1
       fetch : https://nodejs.org/dist/v10.14.1/node-v10.14.1-darwin-x64.tar.gz
   installed : v10.14.1 to /Users/john/deleteme/bin/node
      active : v10.14.2 at /usr/local/bin/node

````

## Alternatives

I considered adding a `--force` and/or adding a message when skipping the install, but I think always doing the install is simpler and clearer.

Could test both the node and npm versions before skipping the install, but again I prefer keeping it simple.

[...]

PR #842, opened 2025-10-29
Lazy updates

sigJoe:
# Pull Request

## Problem

When using `n` to install Node.js versions without an exact version match (e.g., `n 18`, `n lts`, `n auto`), the tool always jumps to the latest version even if the currently active version technically satisfies the requirement. This is inefficient in scenarios like CI/CD pipelines where the existing version may be "good enough". 

If a custom pipeline docker image has `22.21.0` preinstalled, I don't need every deployment (in the case of engines `22`) to update to `22.21.1`. However, I do want `n auto` to support repositories that require alternate Node versions. 

Additionally, developers often only specify a minimum supported release (e.g `>=20`) but then a new Node release (e.g. 25) comes out and breaks everything. For those situations, a lazy update procedure would be convenient such that we can delay fixing Node 25 issues until we're ready to switch to Node 26 LTS.

## Solution

Added a `--lazy` flag that checks if the currently active Node.js version satisfies the requested version or range before installing. If the current version is "good enough", the installation is skipped. Using targets `latest` or `current` will ignore `--lazy`.

The lazy check supports:
- Major.minor ranges (e.g. `4.9` matches any `4.9.x`)
- Major version ranges (e.g. `4` matches any `4.x.x`)
- Release codename as major version ranges (e.g. `argon` matches `4.x.x`)
- Semver operators (e.g. `>=`, `>`, `~`, `^`, `=`)
- Complex ranges via `npx semver` (e.g. `>=8 <12`, `8 || 10`)

## ChangeLog

Added `--lazy` flag to skip installation if current Node.js version satisfies the requested version or range

### DIFF
--- a/bin/n
+++ b/bin/n
@@ -1233,13 +1233,8 @@ function version_satisfies_target() {
   current="${current#v}"
   target="${target#v}"
   
-  # Exact match
-  if [[ "${current}" == "${target}" ]]; then
-    return 0
-  fi
-  
-  # Wildcard
-  if [[ "*" == "${target}" ]]; then
+  # Exact match or wildcard
+  if [[ "$current" == "$target" || $target == "*" ]]; then
     return 0
   fi
   
@@ -1249,7 +1244,7 @@ function version_satisfies_target() {
     local version="${BASH_REMATCH[2]}"
 
     case "${operator}" in
-      '')
+      '' | =)
         # e.g. "22" matches 22.x.x, "22.21" matches 22.21.x, etc.
         if [[ "${version}" =~ ^[0-9]+$ ]]; then
           [[ "${current%%.*}" == "${version}" ]]
@@ -1258,9 +1253,6 @@ function version_satisfies_target() {
         else
           [[ "${current}" == "${version}" ]]
         fi
-        ;;
-      =)
-        [[ "${current}" == "${version}" ]]
         ;;
     \>) version_gte "${current}" "${version}" && [[ "${current}" != "${version}" ]] ;;
     \>=) version_gte "${current}" "${version}" ;;
--- a/bin/n
+++ b/bin/n
@@ -1234,11 +1234,9 @@ function version_satisfies_target() {
   # Remove 'v' prefix if present
   current="${current#v}"
   target="${target#v}"
-  
-  # Exact match or wildcard
-  if [[ "$current" == "$target" || $target == "*" ]]; then
-    return 0
-  fi
+
+  [[ -z "${target}" ]] && return 1
+  [[ "${current}" == "${target}" || "${target}" == "*" ]] && return 0
   
   # Check for operator-based range (only supports what get_package_engine_version does)
   if [[ "${target}" =~ ^([>~^=]|\>\=)?v?([0-9]+(\.[0-9]+){0,2})(.[xX*])?$ ]]; then
@@ -1335,11 +1333,14 @@ function get_latest_resolved_version() {
   fi
 
   if [[ -z "${g_if_needed_node}" ]]; then
-    if ! is_numeric_version "${version}" && [[ "${version}" != "latest" && "${version}" != "current" ]]; then
+    if is_numeric_version "${version}"; then
+      g_if_needed_node="${version}"
+    elif is_lts_codename "${version}" || [[ "${version}" = "lts" ]]; then
       # Codenames (e.g. "argon") will only if-needed match major rather than latest minor/patch
       g_if_needed_node=$(display_major_version "${g_target_node}")
     else
-      g_if_needed_node="${version}"
+      # Other strings (e.g. latest, nightly, rc) should never skip updates
+      g_if_needed_node=""
     fi
   fi
 }
--- a/bin/n
+++ b/bin/n
@@ -789,7 +789,6 @@ install() {
   if [[ "$LAZY" == "true" ]]; then
     if [[ -n "${current_version}" ]]; then
       verbose_log "lazy" "${g_lazy_node}"
-      verbose_log "lazy2" "${g_lazy_node:=$version}"
       if version_satisfies_target "${current_version}" "${g_lazy_node}"; then
         echo "Current version ${current_version} satisfies requirement ${g_lazy_node}"
         return 0
@@ -1336,10 +1335,6 @@ function get_latest_resolved_version() {
     version="${g_target_node}"
   fi
 
-  if is_numeric_version "${version}"; then
-    g_lazy_node="${version}"
-  fi
-
   simple_version=${version#node/} # Only place supporting node/ [sic]
   if is_exact_numeric_version "${simple_version}"; then
     # Just numbers, already resolved, no need to lookup first.
@@ -1350,6 +1345,15 @@ function get_latest_resolved_version() {
   else
     # Complicated recognising exact version, KISS and lookup.
     g_target_node=$(N_MAX_REMOTE_MATCHES=1 display_remote_versions "$version")
+  fi
+
+  if [[ -z "${g_lazy_node}" ]]; then
+    if ! is_numeric_version "${version}" && [[ "${version}" != "latest" && "${version}" != "current" ]]; then
+      # Codenames (e.g. "argon") will only lazy match major rather than latest minor/patch
+      g_lazy_node=$(display_major_version "${g_target_node}")
+    else
+      g_lazy_node="${version}"
+    fi
   fi
 }
 
--- a/bin/n
+++ b/bin/n
@@ -1339,7 +1339,7 @@ function get_latest_resolved_version() {
       # Codenames (e.g. "argon") will only if-needed match major rather than latest minor/patch
       g_if_needed_node=$(display_major_version "${g_target_node}")
     elif [[ "${version}" = "lts" ]]; then
-      g_if_needed_node=">=$(get_latest_lts_lowest_version)"
+      g_if_needed_node="^$(get_latest_lts_lowest_version)"
     else
       # Other strings (e.g. latest, nightly, rc) should never skip updates
       g_if_needed_node=""
--- a/bin/n
+++ b/bin/n
@@ -132,8 +132,12 @@ g_active_node=
 # set by various lookups to allow mixed logging and return value from function, especially for engine and node
 g_target_node=
 
+# Like g_target_node, but retained granularity differs based on context
+g_lazy_node=
+
 DOWNLOAD=false # set to opt-out of activate (install), and opt-in to download (run, exec)
 CLEANUP=false # remove cached download after install
+LAZY=false # check if current version satisfies requirement before installing
 ARCH="${N_ARCH-}"
 SHOW_VERBOSE_LOG="true"
 OFFLINE=false
@@ -402,6 +406,7 @@ Options:
   --cleanup             Remove cached version after install
   -a, --arch            Override system architecture
   --offline             Resolve target version against cached downloads instead of internet lookup
+  --lazy                Skip install if current version already "good enough"
   --all                 ls-remote displays all matches instead of last 20
   --insecure            Turn off certificate checking for https requests (may be needed from behind a proxy server)
   --use-xz/--no-use-xz  Override automatic detection of xz support and enable/disable use of xz compressed node downloads.
@@ -778,6 +783,20 @@ install() {
   get_latest_resolved_version "$1" || return 2
   version="${g_target_node}"
   [[ -n "${version}" ]] || abort "no version found for '$1'"
+  
+  local current_version="$(node --version 2>/dev/null)"
+  verbose_log "current" "${current_version}"
+  if [[ "$LAZY" == "true" ]]; then
+    if [[ -n "${current_version}" ]]; then
+      verbose_log "lazy" "${g_lazy_node}"
+      verbose_log "lazy2" "${g_lazy_node:=$version}"
+      if version_satisfies_target "${current_version}" "${g_lazy_node}"; then
+        echo "Current version ${current_version} satisfies requirement ${g_lazy_node}"
+        return 0
+      fi
+    fi
+  fi
+  
   update_mirror_settings_for_version "$1"
   update_xz_settings_for_version "${version}"
   update_arch_settings_for_version "${version}"
@@ -1106,6 +1125,7 @@ function get_file_node_version() {
 
 function get_package_engine_version() {
   g_target_node=
+  g_lazy_node=
   local filepath="$1"
   verbose_log "found" "${filepath}"
   local range
@@ -1118,6 +1138,7 @@ function get_package_engine_version() {
   fi
   verbose_log "read" "${range}"
   [[ -n "${range}" ]] || return 2
+  g_lazy_node="${range}" # Store range before we start replacing numbers with "current"
   if [[ "*" == "${range}" ]]; then
     verbose_log "target" "current"
     g_target_node="current"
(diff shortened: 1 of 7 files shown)


bin/n
```diff
@@ -1195,6 +1215,78 @@ function get_engine_version() {
...
+          local current_major_minor="${current%.*}"
+          [[ "${current_major_minor}" == "${version}" ]]
+        else
+          [[ "${current}" == "${version}" ]]
+        fi
+        ;;
+      =)
+        [[ "${current}" == "${version}" ]]
+        ;;
+      \>) [[ "$(printf '%s\n%s' "${current}" "${version}" | sort -V | tail -n1)" == "${current}" && "${current}" != "${version}" ]] ;;
```
sigJoe: Technically most of these (apart from `''`) can be handled by `npx semver` logic, but `get_package_engine_version` handled these same operators without `npx semver` so it seemed reasonable for my addition to also handle them without the need for `npx semver`

test/tests/install-if-needed.bats
```diff
@@ -0,0 +1,222 @@
...
+  
+  output="$(node --version)"
+  assert_equal "${output}" "v8.16.0"
+}
+
+
+@test "n --lazy with engine || match" {
+  mkdir -p "${TMP_PREFIX_DIR}/project"
+  cd "${TMP_PREFIX_DIR}/project"
+  echo '{"engines": {"node": "8 || 10"}}' > package.json
```
sigJoe, line 196: Most of the tests use 4.9.1 because I copied style from `install-options`, but these "complex" engine tests used newer releases because they need `npx` and it wasn't available in Node 4.

shadowspawn:
Some past issues related to matching cached installs: #252 #727

bin/n
```diff
@@ -778,6 +783,19 @@ install() {
   get_latest_resolved_version "$1" || return 2
   version="${g_target_node}"
   [[ -n "${version}" ]] || abort "no version found for '$1'"
+  
+  local current_version="$(node --version 2>/dev/null)"
+  verbose_log "current" "${current_version}"
```
shadowspawn, line 797: I noticed this extra log line. Related: #825

shadowspawn:
Version 5 of `n` changed the install behaviour:

> reinstalling active node version always does reinstall (previously silently did nothing)

I can see this option allowing an "optimised" install that skips if the existing version is already the same.

I found one related issue (#574), and the issue before changing the behaviour was #543.

shadowspawn:
I am concerned this adds a bunch of gnarly code for something that has not been much requested, and makes the result of the install depend on previous installs. However, it is opt-in and skips a potentially unnecessary install.

Still thinking about this.

I think `--lazy` is not right. I think the wording could focus on making the install conditional, checking before installing, or skipping install if already compatible. Perhaps `--check-existing`.

sigJoe:
> I am concerned this adds a bunch of gnarly code

The `version_satisfies_target()` could definitely be a lot less gnarly if I relied on `semver` more. The thing that I kept getting stuck on is that installing semver every run only feels marginally better than installing Node every run. Maybe a compromise would be to move `~` and `^` handling (messiest code IMO) under `semver` and keeping `=`, `>` and `>=` as they are?

> I think `--lazy` is not right. I think the wording could focus on making the install conditional, checking before installing, or skipping install if already compatible. Perhaps `--check-existing`.

To clarify, my thinking was along the lines of "lazy loading" by only doing something if/as needed. My original idea was `--if-needed`, although I kept getting stuck on what to name related variables

shadowspawn:
`--if-needed` is good for CLI. Trying some option names:

```console
n --if-needed auto
n --if-needed install 20

n  --check-exisiting auto
n  --check-exisiting install 20

n --skip-existing auto
n --skip-existing install 20
```

sigJoe:
I updated `--lazy` to `--if-needed` and prettified my code additions. 

Also, I updated the Dockerfiles to install libatomic1 for tests that need to install latest (Node 25) and were failing with `node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory`

bin/n
shadowspawn: Yes, the goal in `get_package_engine_version`  was to handle the easy ones directly without needing to invoke semver.

shadowspawn:
This PR includes fixes for node 25 requiring libatomic1, per above comment. Example error from running tests without changes:

```console
 ✗ n latest
   (in test file /mnt/test/tests/install-versions.bats, line 37)
     `output="$(node --version)"' failed with status 127
     installing : node-v25.1.0
          mkdir : /tmp/tmp.Dm9ebsPmxR/n/versions/node/25.1.0
          fetch : https://nodejs.org/dist/v25.1.0/node-v25.1.0-linux-arm64.tar.gz
        copying : node/25.1.0
   node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
   /tmp/tmp.Dm9ebsPmxR/bin/node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
      installed :  (with npm )
   node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory
```

bin/n
```diff
@@ -198,6 +202,13 @@ display_major_version() {
     echo "${version}"
 }
 
+#
+# Synposis: return if a $1 >= $2, given values like x.y.z
+#
+version_gte() {
+  [[ "$(printf '%s\n%s' "$1" "$2" | sort -V | tail -n1)" == "$1" ]]
```
shadowspawn: Conveniently the recent refactor has localised the use of `sort -V`.

Unfortunately `sort -V` is not included in the posix definition of `sort`, and some implementations do not have it. A now fairly historical example was macOS Sierra (#502, #504).

sigJoe: Good catch! I've now borrowed your `| sort  -k1,1n -k2,2n -k3,3n -k4,4n -t .` pattern from 504

bin/n
```diff
@@ -1351,6 +1346,15 @@ function get_latest_resolved_version() {
     # Complicated recognising exact version, KISS and lookup.
     g_target_node=$(N_MAX_REMOTE_MATCHES=1 display_remote_versions "$version")
   fi
+
+  if [[ -z "${g_lazy_node}" ]]; then
+    if ! is_numeric_version "${version}" && [[ "${version}" != "latest" && "${version}" != "current" ]]; then
```
shadowspawn: This test would be better inclusive to just the ones that major match is appropriate for.

bin/n
```diff
@@ -1195,6 +1222,60 @@ function get_engine_version() {
...
+  # Remove 'v' prefix if present
+  current="${current#v}"
+  target="${target#v}"
+  
+  # Exact match or wildcard
+  if [[ "$current" == "$target" || $target == "*" ]]; then
+    return 0
+  fi
+  
+  # Check for operator-based range (only supports what get_package_engine_version does)
```
shadowspawn, line 1241: This code closely follows get_package_engine_version which only has to deal with things that are in engines fields. You might be assuming that everything is a range by now, but `latest` and `current` are currently coming through as strings. This triggers the semver call which does not recognise the label.

bin/n
```diff
@@ -1195,6 +1222,60 @@ function get_engine_version() {
...
+        version_gte "${current}" "${version}" && [[ "${current%.*}" == "${version}" ]]
+      else
+        version_gte "${current}" "${version}" && [[ "${current%%.*}" == "${version}" ]]
+      fi
+      ;;
+    esac
+  else
+    # Complex range, use npx semver for checking
+    command -v npx &> /dev/null || return 1
+    npx --quiet semver@7 "${current}" -r "${target}" &> /dev/null
```
shadowspawn: You left out the `npm_config_yes=true` which means npx blocks waiting for confirmation of the download of semver, and the prompt is hidden by the redirect.

shadowspawn:
The processing for `g_if_needed_node` is currently quite simple. It _mostly_ works by matching major version for a bunch of the labels.

I am wondering if a major version match is appropriate for `lts`. It kinda works, but technically the early versions in the major were not its, and there can be more than one major. A proper match is probably too complicated. (e.g. `ls-remote` only shows the latest version.)

The special case that falls down currently are the stream labels, so might make a call like `n install nightly/26.0.0-nightly202510`. These do get handled by `install` and `ls-remote`. Not as simple as a major version match and perhaps just match exactly.

bin/n
```diff
@@ -1252,6 +1333,15 @@ function get_latest_resolved_version() {
     # Complicated recognising exact version, KISS and lookup.
     g_target_node=$(N_MAX_REMOTE_MATCHES=1 display_remote_versions "$version")
   fi
+
+  if [[ -z "${g_if_needed_node}" ]]; then
+    if ! is_numeric_version "${version}" && [[ "${version}" != "latest" && "${version}" != "current" ]]; then
+      # Codenames (e.g. "argon") will only if-needed match major rather than latest minor/patch
+      g_if_needed_node=$(display_major_version "${g_target_node}")
+    else
+      g_if_needed_node="${version}"
```
shadowspawn: This works nicely for numeric partial versions like `10.2`. However if you are assuming you have reduced everything to numeric ranges, you might need another fallback to set it to `g_target_node` for unsupported non-numeric cases (like say `nightly`).

bin/n
```diff
@@ -1195,6 +1222,60 @@ function get_engine_version() {
...
+  fi
+  
+  # Check for operator-based range (only supports what get_package_engine_version does)
+  if [[ "${target}" =~ ^([>~^=]|\>\=)?v?([0-9]+(\.[0-9]+){0,2})(.[xX*])?$ ]]; then
+    local operator="${BASH_REMATCH[1]}"
+    local version="${BASH_REMATCH[2]}"
+
+    case "${operator}" in
+      '' | =)
+        # e.g. "22" matches 22.x.x, "22.21" matches 22.21.x, etc.
```
shadowspawn, line 1248: For interest, the way partial numeric matches are handled in `display_local_versions` and `display_remote_versions`  is to look for a matching prefix. For example:

https://github.com/tj/n/blob/f52d2172f12cd76f0efe9524690723f52ab74f40/bin/n#L1310-L1315

shadowspawn: (Would be about as much code and still complex, so not suggesting change.)

shadowspawn:
I eventually got annoyed with doing full installs to test the lookup code and added `N_TEST_DISPLAY_LATEST_RESOLVED_VERSION` so could test lots of cases cheaply. We could so something similar to allow lighter weight  `--if_needed` testing. 

Unless you want it now for faster testing of PR, feel free to leave that for me or at least until after I have decided whether likely to accept PR.

bin/n
sigJoe: I don't know of a way to implement that effectively... Node keeps adding new codenames over time and we've already got `krypton or jod or iron or hydrogen or gallium or fermium or erbium or dubnium or carbon or boron or argon`

If you you think `latest` and `current` should major match as well, then I could simplify to `is_alphabetic()` or something like that

Update: disregard, I just noticed that `is_lts_codename()` exists.

sigJoe: Per your later comments, I did totally overlook the idea of nightly and rc and so on... I'll need to think on this

shadowspawn:
Light weight version checking proof of concept.

Add `N_TEST_DISPLAY_IF_NEEDED` after `N_TEST_DISPLAY_LATEST_RESOLVED_VERSION`:

```bash
    N_TEST_DISPLAY_LATEST_RESOLVED_VERSION) shift; get_latest_resolved_version "$1" > /dev/null || exit 2; echo "${g_target_node}"; exit ;;
    N_TEST_DISPLAY_IF_NEEDED) shift; display_if_needed "$1" "$2"; exit ;;
```

```bash
#
# display_if_needed target_version active_version
#
display_if_needed() {
  local active_version="$1"
  local target_version="$2"
  CHECK_IF_NEEDED="TRUE"

  # calculate g_if_needed_node as side-effect
  get_latest_resolved_version "${target_version}" || exit 2

  if version_satisfies_target "${active_version}" "${g_if_needed_node}"; then
    echo "install not needed: active version ${active_version} satisfies requirement ${g_if_needed_node}"
  else
    echo "install needed: active version ${active_version} does not satisfy requirement ${g_if_needed_node}"
  fi
}
```

```console
% n N_TEST_DISPLAY_IF_NEEDED 8.1.2 carbon
install not needed: active version 8.1.2 satisfies requirement 8
% n N_TEST_DISPLAY_IF_NEEDED 8.1.2 10
install needed: active version 8.1.2 does not satisfy requirement 10
```

sigJoe:
> I am wondering if a major version match is appropriate for `lts`. It kinda works, but technically the early versions in the major were not its, and there can be more than one major.

I went with major matching for `lts` in [0ca1037](https://github.com/tj/n/pull/842/commits/0ca1037dc730344d53ab9683a38ca85321fe31f0) but then reconsidered and added `get_latest_lts_lowest_version` in [4e8073b](https://github.com/tj/n/pull/842/commits/4e8073b511a0cd5268277fb29c5b8749ec39e7de) which more precisely determines the requirement.

> The special case that falls down currently are the stream labels, so might make a call like `n install nightly/26.0.0-nightly202510`. These do get handled by `install` and `ls-remote`. Not as simple as a major version match and perhaps just match exactly.

New behaviour is to default to `g_if_needed_node=""` in unhandled scenarios (e.g. nightly) which will force a download/install.

bin/n
```diff
@@ -1252,6 +1331,20 @@ function get_latest_resolved_version() {
...
   fi
+
+  if [[ -z "${g_if_needed_node}" ]]; then
+    if is_numeric_version "${version}"; then
+      g_if_needed_node="${version}"
+    elif is_lts_codename "${version}"; then
+      # Codenames (e.g. "argon") will only if-needed match major rather than latest minor/patch
+      g_if_needed_node=$(display_major_version "${g_target_node}")
+    elif [[ "${version}" = "lts" ]]; then
+      g_if_needed_node=">=$(get_latest_lts_lowest_version)"
```
shadowspawn: Nice try excluding the pre-LTS minor versions. Still need to match the major version though, or allow newer odd non-LTS versions.

```console
% n current
     current : v25.2.0
     copying : node/25.2.0
   installed : v25.2.0 (with npm 11.6.2)
% n --if-needed lts
     current : v25.2.0
   if-needed : >=24.11.0
```

bin/n
```diff
@@ -1252,6 +1331,20 @@ function get_latest_resolved_version() {
...
+    else
+      # Other strings (e.g. latest, nightly, rc) should never skip updates
+      g_if_needed_node=""
```
shadowspawn: What I had suggested here was `g_target_node` so still skip exact matches. Did you have some targets in mind that would get false positives? You might have been worried about the exotic targets `nightly` and `rc` but they have metadata after the version number so not just numeric version.

```suggestion
      g_if_needed_node="${g_target_node}"
```

With this change:
```console
% n nightly
     current : v25.0.0-rc.2
     copying : nightly/26.0.0-nightly202511139079d3fd50
   installed : v26.0.0-nightly202511139079d3fd50 (with npm 11.6.2)
% n --if-needed nightly
     current : v26.0.0-nightly202511139079d3fd50
   if-needed : 26.0.0-nightly202511139079d3fd50
Current version v26.0.0-nightly202511139079d3fd50 satisfies requirement 26.0.0-nightly202511139079d3fd50
```

sigJoe: No, I didn't have any specific ideas. I just didn't consider exact matches would matter if it said "satisfies requirement..." or if it was handled by the pre-existing logic.

sigJoe: Not that I disagree with your comment, but my previous experience with `n` was relatively superficial and we're now well outside the areas of that I thought I was familiar with.

This was supposed to just be a quick 20 minute adventure :)

shadowspawn: Don't feel you have to keep considering exotic targets and weird POSIX restrictions if your end-goal is a proof of concept. It is pretty good now!

Release quality is a higher bar. Adding solid new features can be tricky because there are lots of interactions to consider with what is already supported.

bin/n
```diff
@@ -1339,7 +1339,7 @@ function get_latest_resolved_version() {
       # Codenames (e.g. "argon") will only if-needed match major rather than latest minor/patch
       g_if_needed_node=$(display_major_version "${g_target_node}")
     elif [[ "${version}" = "lts" ]]; then
-      g_if_needed_node=">=$(get_latest_lts_lowest_version)"
+      g_if_needed_node="^$(get_latest_lts_lowest_version)"
```
shadowspawn, line 1342: Nice use of existing support, I had not thought of that.

shadowspawn:
I am going to use this as basis for a couple of pieces of work: 
- fixing the libatomic1 in containers for node 25
- hopefully finish off `--if-needed`. check for interactions and consider how gnarly code is (mostly because of `engine`, which starts gnarly anyway)

I will add @sigJoe as co-author on both. Thanks for the suggestion and for the work on refining behaviour @sigJoe !

Issue #847, opened 2026-02-19
Revise labels for new Node.js Release Schedule


shadowspawn:
The Node.js release schedule is changing. The announcement is:
- https://nodejs-org-git-fork-ulisesgascon-release-announcement-openjs.vercel.app/en/blog/announcements/evolving-the-nodejs-release-schedule

Old Node.js terminology (https://nodejs.org/en/about/previous-releases):
- even: current / active LTS / maintenance LTS
- odd: current / maintenance

New terminology used in announcement:
- alpha / interim / LTS

I think this makes the little used node support labels somewhat obsolete (like `lts_active`), and `current` is not mentioned. `lts_latest` got used in some training material and will retain that as an alias even if do not document. Likewise, stable.

Not sure if `n` users will want to target the latest version _after_ it reaches alpha? `n` uses the index information from https://github.com/nodejs/nodejs-dist-indexer and does not do date calculations. I am watching that repo to see if alpha/interim get added to the index.

shadowspawn:
Another blog entry coming: https://github.com/nodejs/nodejs.org/pull/8631

shadowspawn:
https://nodejs.org/en/blog/announcements/evolving-the-nodejs-release-schedule

Terminology used in announcement:

- alpha / current / LTS

shadowspawn:
My current thinking is to modify existing `current` to allow avoiding alphas:
- `latest`: highest release, may be alpha or current/active depending on timing
- `current`: highest release not alpha, may be current/active or LTS depending on timing
- `lts`: highest LTS release

I was concerned about the "new" alpha stage, but it isn't that different to the odd numbered releases.

 Likewise I wonder about a warning for alpha versions, but it will be visible in the version:
```console
$ n latest
     copying : node/27.0.0-alpha.12
     ...
$ n current
     copying : node/26.2.0
     ...
```

PR #850, opened 2026-08-15
Add support for alpha versions of node

shadowspawn:
See #847 for more background and discussion.

## Problem

Node.js is switching to yearly releases, with alpha -> current -> lts

What do we do with labels to support new policy?

## Solution

Keep using `latest` and `current` labels, with refined behaviour for `current`:

- `latest`: newest release
- `current`: newest release ignoring alpha versions

Node.js will have periods when there is not a version in its "current" phase. We just match our `current` label to newest non-alpha, so may actually be in current or LTS phase. This seems more useful than being pedantic and returning an error.

Add some extra handling for:

- excluding alpha versions from `current`
- sorting local listing so prerelease comes before release versions

Remove `lts_latest` from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,20 +10,20 @@ Node.js version management: no subshells, no profile setup, no convoluted API, j
 ![usage animation](https://nimit.io/images/n/n.gif)
 
 - [`n` – Interactively Manage Your Node.js Versions](#n--interactively-manage-your-nodejs-versions)
-    - [Supported Platforms](#supported-platforms)
-    - [Installation](#installation)
-        - [Third Party Installers](#third-party-installers)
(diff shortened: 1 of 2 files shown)


shadowspawn:
Node.js is going with triple alpha numbers so can show semver major changes. First release going to be 27.0.0-alpha.0.0.0.

- https://github.com/nodejs/Release/issues/1154
- https://github.com/nodejs/node/pull/63135

beceqada32ps3 (approved):
- #851 

- [x] Support devEngines
- [x] Support the guidelines
#851

beceqada32ps3 (changes requested):
Where gonna close it soon

